PR #12, opened 2016-09-19
Introduce phar building

nervo:


.gitignore
```diff
@@ -1,2 +1,3 @@
-vendor
-composer.lock
+/vendor/
+/composer.lock
```
ogizanagi, line 2: Out-of-topic, but as it's highlighted here: 
As it's more an application than a library, why would you ignore the `composer.lock` file ?

ogizanagi:
Damn. This `box.json` file is pretty accurate 😮 😅

.gitignore
chalasr: > it's more an application than a library

Is there a reason for this thought?

I would ignore the `composer.lock` too, assuming it is indeed a library (as stated in the `composer.json`) and considering we don't need to lock deps at point releases

ogizanagi: To me, manala looks like a full fledged application, ready to use. Not as a library your project might depend on.
After all, that's why you're building an executable phar. It represents a runnable application/tool, more than a library you'll require and use in your code.

`composer/composer` for instance (despite the fact the `composer.json` file describes it as a `library`...) locks its dependencies.

nervo: I'm agree too that it should be considered as a library. @ogizanagi what would be the benefits of having the lock file in the repo ?

ogizanagi: (BTW the [`type`](https://getcomposer.org/doc/04-schema.md#type) is currently missing from the [`composer.json` file](https://github.com/manala/manalize/blob/master/composer.json) ^^'

nervo: Well, i'm pretty convinced :) @chalasr it's up to you

nervo: @ogizanagi yes, but not the "bin" one :)

ogizanagi: No problem to me anyway. It was just out of curiosity :)

chalasr: > Not as a library your project might depend on.

Not sure to agree. Should php-cs-fixer, gulp, and such tools be considered as applications rather than libraries so? 
As for these packages, there is great chances for that a dotfile will be added to projects using this project

> (BTW the type is currently missing from the composer.json file ^^'

Thank you ☺️

chalasr: Even if I totally understand your thought about the difference between such dependency and whatever package you'll directly use in your code

ogizanagi: - [`php-cs-fixer` is an application](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/composer.json#L3) but does not lock deps.
- [`composer/composer`](https://github.com/composer/composer/blob/master/composer.json#L6) is a library despite the fact I consider it a project/application according to the criteria mentioned above.
- [`phpunit`](https://github.com/sebastianbergmann/phpunit/blob/5.5/composer.json#L4) is a library. But that's fine to me because it's a tool and a unit testing framework you use in your codebase. It does not lock deps.

It actually seems there is no strong consensus on what is a project, an application, a tool, or a library in the PHP community, and where deps should be locked or not. Let's not find one here :trollface:

chalasr: Sure, sorry for sticking 😅  
I just never considered this and (out of pure curiosity) I wanted to get more about when should a third party tool be considered as an application. My php-cs-fixer argument being returned against me, I will reconsider the question :)

About the lock file, let's ignore it for now, if it becomes a real need a day we'll advise

box.json
```diff
@@ -0,0 +1,42 @@
...
+      "exclude": [
+        "Tests",
+        "Tester"
+      ],
+      "in": [
+        "vendor/symfony/console",
+        "vendor/symfony/filesystem",
+        "vendor/symfony/process",
+        "vendor/symfony/stopwatch",
+        "vendor/symfony/polyfill-mbstring"
```
chalasr, line 34: Would it be possible to summarize this list in a `vendor/symfony`?  So we would juste have to update this file when adding packages from a new vendor

nervo: Hm, not so simple.
It will break the global "Test" exclusions, and force you to define it for each vendor. That's how finder works...
But maybe there's a trick with finder, i don't know.
Btw, this is how composer does.

chalasr: Let's keep it as is then, not a big deal.

.gitignore
chalasr: Last thought for edge case: it would be terrible to see someone installing this as local dependency for its project to be bothered by conflicts when trying to require some other packages, because of our tilde-bounded constraints (e.g. `symfony/console:~3.1`). ~~I hope the `application` type helps in that way~~ What could composer do against that...    be flexible on our constraints is the better we can do.

ogizanagi: It appears to me there is no reason this project would be installed as a local dependency.
However, you made a point with tightly bounded constraints, in case this project is installed globally, as people can do with `phpunit`, `psysh`, `cs-fixer`, ...

Reviewing the current constraints, not locking the deps and configuring the Travis build matrix in order to test lowest and highest dep constraints would help ensuring that, if we're actually going this way. 
But is it actually worth it ? I don't think so, because the project is mainly thought to be used through the `.phar` as a standalone application, with its own dep constraints, and so... :trollface: => 🚪 

(This debate is endless 😸 )

PR #13, opened 2016-09-19
Manala\Manalize => Manala

chalasr:
Objection? :)


nervo:
Dude, we should also rename the repo itself from manalize to manala, no ?

nervo:
(but right now, it's :smoking: time )

chalasr:
> we should also rename the repo itself

Indeed :) I plan to do it just before merging this one

PR #15, opened 2016-09-21
Fix usage through phar://

chalasr:


bin/manala
```diff
@@ -15,6 +15,8 @@ require_once __DIR__.'/../vendor/autoload.php';
 use Manala\Command;
 use Symfony\Component\Console\Application;
 
+define('MANALA_DIR', \Phar::running() ?: __DIR__.'/../');
```
ogizanagi: Out of curiosity: is the `\Phar::running` call needed ? Won't the path always be `__DIR__.'/../'` whether you are running the phar or the php script ?

chalasr: Indeed, it was some debugging leavings. Removed

bin/manala
```diff
@@ -15,8 +15,6 @@ require_once __DIR__.'/../vendor/autoload.php';
 use Manala\Command;
 use Symfony\Component\Console\Application;
 
-define('MANALA_DIR', \Phar::running() ?: __DIR__.'/../');
```
ogizanagi: I would have kept the constant, though

ogizanagi (approved):

src/Config/Renderer.php
```diff
@@ -31,7 +31,7 @@ public static function render(Config $config, Vars $vars)
     {
         $template = $config->getTemplate();
 
-        if (!is_writable($template) || !is_readable($template)) {
+        if (!is_readable($template)) {
             throw new \RuntimeException(sprintf('The template file "%s" is either not readable, not writable or it doesn\'t exist.', $template));
```
benji07: you should change the error message, now that you only check if the template is readable

benji07 (changes requested):

bin/manala
chalasr: Do you just find it convenient? We can live without as there is only one occurrence of `__DIR__` in `src/` for now (should not growth).

src/Config/Renderer.php
chalasr: Thanks, fixed

bin/manala
ogizanagi: Sure, that's only for convenient reasons (and preserve from any issue if the files requiring the get the project root are moved elsewhere).

But that's not a big deal actually :)

PR #18, opened 2016-09-23
Remove dead code & perform static code analysis

ogizanagi:
Those are only suggestions, if one of them does not please you, I'll revert the related changes.

### DIFF
--- a/src/Command/Setup.php
+++ b/src/Command/Setup.php
@@ -57,7 +57,10 @@ class Setup extends Command
         $io->setDecorated(true);
         $io->comment(sprintf('Composing your <info>%s</info> environment', (string) $envType));
 
-        $vars = new Vars($io->ask('Vendor name', null, [$this, 'validateVar']), $io->ask('App name', null, [$this, 'validateVar']));
+        $vars = new Vars(
+            $io->ask('Vendor name', null, [Vars::class, 'validate']),
+            $io->ask('App name', null, [Vars::class, 'validate'])
+        );
         $env = EnvFactory::createEnv($envType);
 
         foreach (Dumper::dump($env, $vars, $cwd) as $dumpTarget) {
@@ -67,23 +70,5 @@ class Setup extends Command
         $io->success('Environment successfully configured');
 
         return 0;
-    }
-
-    /**
-     * Checks that a given configuration value is properly formatted.
-     *
-     * @param string $value The value to assert
-     *
-     * @return string The validated value
-     *
-     * @throws \InvalidArgumentException If the value is incorrect
-     */
-    public function validateVar($value)
-    {
-        if (!preg_match('/^([-A-Z0-9])*$/i', $value)) {
-            throw new RuntimeException(sprintf('This value must contain only alphanumeric characters and hyphens.', $value));
-        }
-
-        return $value;
     }
 }
--- a/src/Process/Build.php
+++ b/src/Process/Build.php
@@ -37,9 +37,10 @@ class Build
     private $errorOutput;
 
     /**
-     * @param string $cwd
+     * @param string   $cwd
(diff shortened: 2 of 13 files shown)


src/Command/Setup.php
```diff
@@ -57,7 +57,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $io->setDecorated(true);
         $io->comment(sprintf('Composing your <info>%s</info> environment', (string) $envType));
 
-        $vars = new Vars($io->ask('Vendor name', null, [$this, 'validateVar']), $io->ask('App name', null, [$this, 'validateVar']));
+        $vars = new Vars(
+            $io->ask('Vendor name', null, [$this, 'validateVar']),
```
ogizanagi: Calling a constructor multi lined is often better for readability, when the number of arguments exceeds a certain amount, or like in this case, where the arguments are not pre-populated in tmp vars.

src/Command/Setup.php
```diff
@@ -81,7 +84,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
     public function validateVar($value)
     {
         if (!preg_match('/^([-A-Z0-9])*$/i', $value)) {
-            throw new RuntimeException(sprintf('This value must contain only alphanumeric characters and hyphens.', $value));
+            throw new \InvalidArgumentException(sprintf(
```
ogizanagi: Fixed the exception according to the doc block.

But also suggest to use multilined sprintf declaration for exceptions when the exception construct exceeds 120 chars.
Most of the time, it is sufficient to make the php code sniffer happy, and help for readability :)

src/Command/Setup.php
```diff
@@ -81,7 +84,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
     public function validateVar($value)
```
ogizanagi: IMHO, this method does validate vars, indeed, but is specific to the `app` & `vendor` vars. Of course those are the only vars now, but this method thus isn't a generic `validateVar` one for further development, or am I wrong ? :/

Above all, I think it should be moved into a dedicated class, but I failed to find a good candidate. WDYT ? :)

src/Command/Setup.php
chalasr: In this case I'm more like:

``` php
throw new \Exception(
    sprintf(...)
);
```

But I'm fine with yours 👍

src/Command/Setup.php
chalasr: Yes, this method bothers me. Maybe a static `Vars::validate()` or something else? As they should be very only used for these vars.

src/Process/Build.php
```diff
@@ -53,9 +54,11 @@ public function __construct($cwd, $tasks = [])
     /**
      * Starts and returns the running process.
      *
+     * @param callable|null $callback
+     *
      * @return \IteratorAggregate
```
chalasr: Could you fix this one? It doesn't return anything anymore (also maybe you prefer have a return like the last exit code)

src/Command/Setup.php
ogizanagi: The idea behind this syntax is that `sprintf` is not that much important and can stay inlined with the exception constructor call, whereas listing the exception message along with the arguments is visually more effective in order to apprehend the exception message placeholders.

src/Command/Setup.php
ogizanagi: Fine. Simply moved this method in the `Vars` class as suggested :)

src/Process/Build.php
ogizanagi: Fixed (and returned last exit code)

chalasr (approved):

chalasr:
Thank you @ogizanagi

PR #19, opened 2016-09-23
Add Requirement checks command

xavier-rdo:
**Acceptance criteria:**
- [x] Distinguish between requirements & suggestions (1)
- [x] Introduce checking processors that can be mocked (to make tests environment-agnostic) (2)
- [x] Introduce different checker types : binary checker, vagrant plugin checker, etc. (3)
- [x] Required version: semver-formatted (4)
- [x] Unit tests
- [x] Console output colored according to requirement level (required, recommended, etc.)

(1) vagrant & vagrant plugin landrush are required, ansible is suggested (required to deploy, not to provision VM)

(2) Processing a binary != processing a vagrant plugin

(3) Check vagrant & ansible (binaries) but also check landrush (vagrant plugin) : 
- check that a binary is OK: `<binary> --version`
- check that a vagrant plugin is OK: `vagrant plugin list` and check that the expected plugin is in the list

(4) See https://github.com/composer/semver ; Example: "~1.8"

TODO LATER : 
- Functional test (with venv - virtual environments - https://docs.python.org/3/library/venv.html ?)

### DIFF
--- a/src/Command/CheckRequirements.php
+++ b/src/Command/CheckRequirements.php
@@ -1,0 +1,138 @@
+<?php
+
+/*
+ * This file is part of the Manala package.
+ *
+ * (c) Manala <contact@manala.io>
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Command;
+
+use Manala\Config\Requirement\Common\RequirementLevel;
+use Manala\Config\Requirement\Factory\HandlerFactoryResolver;
+use Manala\Config\Requirement\Requirement;
+use Manala\Config\Requirement\RequirementChecker;
+use Manala\Config\Requirement\Violation\RequirementViolation;
+use Manala\Config\Requirement\Violation\RequirementViolationLabelBuilder;
+use Manala\Config\Requirement\Violation\RequirementViolationList;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Formatter\OutputFormatterStyle;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Checks if the host's environment meets Manala's requirements (Vagrant, Ansible, etc.).
+ *
+ * @author Xavier Roldo <xavier.roldo@elao.com>
+ */
+class CheckRequirements extends Command
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('check:requirements')
+            ->setDescription("Checks if your host's environment meets Manala's requirements (Vagrant, Ansible, etc.)")
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->addFormatter($output);
+
+        $violationList = new RequirementViolationList();
+        $requirementChecker = new RequirementChecker(
+            new HandlerFactoryResolver(),
+            new RequirementViolationLabelBuilder()
+        );
+
+        foreach($this->getRequirements() as $requirement) {
+            $output->writeln('Checking ' . $requirement->getName());
+            $requirementChecker->check($requirement, $violationList);
+        }
+
+        if (count($violationList) > 0) {
+            foreach ($violationList as $violation) {
+                $description = $this->getFormattedViolationDescription($violation);
+                $output->writeln($description);
+            }
+        }
+
+        if (!$violationList->containsRequiredViolations()) {
+            $output->writeln('Congratulations ! Everything seems OK.');
+            if ($violationList->containsRecommendedViolations()) {
+                $output->writeln('Yet, some recommendations have been emitted (see above).');
+            }
+        }
+
+    }
+
+    /**
+     * @return Requirement[]
+     */
+    private function getRequirements()
+    {
+        // @TODO: see https://github.com/composer/semver and adopt it for version requirements
+        return [
+            new Requirement(
+                'vagrant',
+                Requirement::TYPE_BINARY,
+                RequirementLevel::REQUIRED,
+                '1.8.4',
+                '>=',
+                'See https://www.vagrantup.com/downloads.html'
+            ),
+            new Requirement(
+                'landrush',
+                Requirement::TYPE_VAGRANT_PLUGIN,
+                RequirementLevel::REQUIRED,
+                '0.18.0',
+                '>=',
+                'See https://github.com/vagrant-landrush/landrush'
+            ),
+            new Requirement(
+                'ansible',
+                Requirement::TYPE_BINARY,
+                RequirementLevel::RECOMMENDED,
+                '2.0.0',
+                '>=',
+                'Required only if you intend to use the deploy role. See http://docs.ansible.com/ansible/intro_installation.html'
+            ),
+        ];
+    }
+
+    /**
+     * @param OutputInterface $output
+     */
+    private function addFormatter(OutputInterface $output)
+    {
+        $errorStyle = new OutputFormatterStyle('red');
+        $output->getFormatter()->setStyle('error', $errorStyle);
+
+        $warningStyle = new OutputFormatterStyle('yellow');
+        $output->getFormatter()->setStyle('warning', $warningStyle);
+    }
+
+    /**
+     * @param RequirementViolation $violation
+     *
+     * @return string Formatted violation description displayed to user.
+     */
+    private function getFormattedViolationDescription(RequirementViolation $violation)
+    {
+        $resultPattern = $violation->isRequired() ? '<error>%s</error>' : '<warning>%s</warning>';
+        $displayedText = $violation->getLabel();
+        $displayedText .= $violation->getHelp() ? ' ' . $violation->getHelp() : '';
+
+        return sprintf($resultPattern, $displayedText);
+    }
+
+}
--- a/src/Command/CheckRequirements.php
+++ b/src/Command/CheckRequirements.php
@@ -80,30 +80,26 @@ class CheckRequirements extends Command
      */
     private function getRequirements()
     {
-        // @TODO: see https://github.com/composer/semver and adopt it for version requirements
         return [
             new Requirement(
                 'vagrant',
                 Requirement::TYPE_BINARY,
                 RequirementLevel::REQUIRED,
-                '1.8.4',
-                '>=',
+                '>= 1.8.0 < 1.8.5 || ^1.8.6', // /!\ Exclude vagrant 1.8.5 (Manala incompatible)
                 'See https://www.vagrantup.com/downloads.html'
             ),
             new Requirement(
                 'landrush',
                 Requirement::TYPE_VAGRANT_PLUGIN,
                 RequirementLevel::REQUIRED,
-                '0.18.0',
-                '>=',
+                '^0.18',
                 'See https://github.com/vagrant-landrush/landrush'
             ),
             new Requirement(
                 'ansible',
                 Requirement::TYPE_BINARY,
                 RequirementLevel::RECOMMENDED,
-                '2.0.0',
-                '>=',
+                '^2.0.0',
                 'Required only if you intend to use the deploy role. See http://docs.ansible.com/ansible/intro_installation.html'
             ),
         ];
--- a/src/Command/CheckRequirements.php
+++ b/src/Command/CheckRequirements.php
@@ -11,14 +11,14 @@
 
 namespace Manala\Command;
 
-use Manala\Config\Requirement\Common\RequirementLevel;
 use Manala\Config\Requirement\Factory\HandlerFactoryResolver;
-use Manala\Config\Requirement\Requirement;
 use Manala\Config\Requirement\RequirementChecker;
+use Manala\Config\Requirement\RequirementRepository;
 use Manala\Config\Requirement\Violation\RequirementViolation;
 use Manala\Config\Requirement\Violation\RequirementViolationLabelBuilder;
 use Manala\Config\Requirement\Violation\RequirementViolationList;
 use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Formatter\OutputFormatterInterface;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
@@ -46,7 +46,7 @@ class CheckRequirements extends Command
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $this->addFormatter($output);
+        $this->configureFormatter($output->getFormatter());
 
         $violationList = new RequirementViolationList();
         $requirementChecker = new RequirementChecker(
@@ -54,8 +54,8 @@ class CheckRequirements extends Command
             new RequirementViolationLabelBuilder()
         );
 
-        foreach($this->getRequirements() as $requirement) {
-            $output->writeln('Checking ' . $requirement->getName());
+        foreach (RequirementRepository::getRequirements() as $requirement) {
+            $output->writeln('Checking '.$requirement->getName());
             $requirementChecker->check($requirement, $violationList);
         }
 
@@ -72,63 +72,31 @@ class CheckRequirements extends Command
                 $output->writeln('Yet, some recommendations have been emitted (see above).');
             }
         }
-
     }
 
     /**
-     * @return Requirement[]
+     * @param OutputFormatterInterface $formatter
      */
-    private function getRequirements()
-    {
-        return [
-            new Requirement(
-                'vagrant',
-                Requirement::TYPE_BINARY,
-                RequirementLevel::REQUIRED,
-                '>= 1.8.0 < 1.8.5 || ^1.8.6', // /!\ Exclude vagrant 1.8.5 (Manala incompatible)
-                'See https://www.vagrantup.com/downloads.html'
-            ),
-            new Requirement(
-                'landrush',
-                Requirement::TYPE_VAGRANT_PLUGIN,
-                RequirementLevel::REQUIRED,
-                '^0.18',
-                'See https://github.com/vagrant-landrush/landrush'
-            ),
-            new Requirement(
-                'ansible',
-                Requirement::TYPE_BINARY,
-                RequirementLevel::RECOMMENDED,
-                '^2.0.0',
-                'Required only if you intend to use the deploy role. See http://docs.ansible.com/ansible/intro_installation.html'
-            ),
-        ];
-    }
-
-    /**
-     * @param OutputInterface $output
-     */
-    private function addFormatter(OutputInterface $output)
+    private function configureFormatter(OutputFormatterInterface $formatter)
     {
         $errorStyle = new OutputFormatterStyle('red');
-        $output->getFormatter()->setStyle('error', $errorStyle);
+        $formatter->setStyle('error', $errorStyle);
 
         $warningStyle = new OutputFormatterStyle('yellow');
-        $output->getFormatter()->setStyle('warning', $warningStyle);
+        $formatter->setStyle('warning', $warningStyle);
     }
 
     /**
      * @param RequirementViolation $violation
      *
-     * @return string Formatted violation description displayed to user.
+     * @return string Formatted violation description displayed to user
      */
     private function getFormattedViolationDescription(RequirementViolation $violation)
     {
         $resultPattern = $violation->isRequired() ? '<error>%s</error>' : '<warning>%s</warning>';
         $displayedText = $violation->getLabel();
-        $displayedText .= $violation->getHelp() ? ' ' . $violation->getHelp() : '';
+        $displayedText .= $violation->getHelp() ? ' '.$violation->getHelp() : '';
 
         return sprintf($resultPattern, $displayedText);
     }
-
 }
--- a/src/Config/Requirement/Common/RequirementLevel.php
+++ b/src/Config/Requirement/Common/RequirementLevel.php
@@ -1,0 +1,18 @@
+<?php
+
+/*
+ * This file is part of the Manala package.
+ *
+ * (c) Manala <contact@manala.io>
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Config\Requirement\Common;
+
+class RequirementLevel
+{
+    const REQUIRED = 1;
+    const RECOMMENDED = 2;
+}
--- a/src/Config/Requirement/Common/RequirementLevel.php
+++ b/src/Config/Requirement/Common/RequirementLevel.php
@@ -11,7 +11,7 @@
 
 namespace Manala\Config\Requirement\Common;
 
-class RequirementLevel
+final class RequirementLevel
 {
     const REQUIRED = 1;
     const RECOMMENDED = 2;
--- a/src/Config/Requirement/Common/RequirementLevelHolderTrait.php
+++ b/src/Config/Requirement/Common/RequirementLevelHolderTrait.php
@@ -1,0 +1,51 @@
+<?php
+
+/*
+ * This file is part of the Manala package.
+ *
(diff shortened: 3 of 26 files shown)


src/Command/CheckRequirements.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+class CheckRequirements extends Command
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('check_requirements')
```
ogizanagi: I'll suggest `requirements:check` instead (as the `security:check` command provided by the `sensiolabs/security-checker`)

src/Config/Requirement/RequirementViolationList.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * (c) Manala <contact@manala.io>
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Config\Requirement;
+
+class RequirementViolationList
```
ogizanagi: Would be great if this class was iterable :)

ogizanagi: And authorize array access ! :D

chalasr: 👍  but let's let @xavier-rdo do its magic and come back when `[WIP]` is no more in the title :)

ogizanagi: He works better under pressure :trollface:

chalasr: haha

src/Command/CheckRequirements.php
```diff
@@ -0,0 +1,97 @@
...
+use Manala\Config\Requirement\RequirementChecker;
+use Manala\Config\Requirement\Violation\RequirementViolationList;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Checks if the host's environment meets Manala's requirements (Vagrant, Ansible, etc.).
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
```
chalasr:
``` diff
- @author Robin Chalas <robin.chalas@gmail.com>
+ @author Xavier Roldo <xavier.roldo@elao.com>
```

On every classes introduced here, so we can easily kill you in the street in case of problem 
:trollface:

ogizanagi: That's exactly why I'd rather use `Elao <contac@elao.com>`. Reviewers do have their part of responsibility. Let's kill everyone if it's not good enough :trollface:

chalasr: The header is here for that (here `Manala <contact@manala.io>`). 
The real reason is just that IMO anyone writing code deserves the credits for :)

src/Config/Requirement/Processor/VagrantPluginProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+
+        return $output;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCommand($name)
+    {
+        return sprintf('vagrant plugin list | grep %s', $name);
```
xavier-rdo: _grep_ : excludes Windows user ;( What's your opinion about that ?

src/Command/CheckRequirements.php
```diff
@@ -0,0 +1,134 @@
...
+                $output->writeln('Yet, some recommendations have been emitted (see above).');
+            }
+        }
+
+    }
+
+    /**
+     * @return Requirement[]
+     */
+    private function getRequirements()
```
xavier-rdo: @TODO: have a dedicated service that instantiates this list of requirements (to ease maintainability, updates) ?

xavier-rdo:
Concerning the @author mention, i will do as you want (I have no opinion about that point).

src/Command/CheckRequirements.php
```diff
@@ -0,0 +1,134 @@
...
+                Requirement::TYPE_BINARY,
+                RequirementLevel::REQUIRED,
+                '>= 1.8.0 < 1.8.5 || ^1.8.6', // /!\ Exclude vagrant 1.8.5 (Manala incompatible)
+                'See https://www.vagrantup.com/downloads.html'
+            ),
+            new Requirement(
+                'landrush',
+                Requirement::TYPE_VAGRANT_PLUGIN,
+                RequirementLevel::REQUIRED,
+                '^0.18',
```
chalasr: This means `0.18.*`, so something like `0.19.0` would not satisfy this requirement. I think we can support all versions between `0.18.0` and `1.0.0`, so using `~0.18` could do the job.

src/Command/CheckRequirements.php
```diff
@@ -0,0 +1,134 @@
...
+                'See https://github.com/vagrant-landrush/landrush'
+            ),
+            new Requirement(
+                'ansible',
+                Requirement::TYPE_BINARY,
+                RequirementLevel::RECOMMENDED,
+                '^2.0.0',
```
chalasr: Could be `^2.0` as well as `landrush` could be `~0.18.0` instead of `~0.18`, so we need to be consistent here, either we explicitly indicate the last `0` (which is not necessary but fine to me) or not.

chalasr:
Very nice job here. I'll do my best to make it fail and be back :)

src/Command/CheckRequirements.php
```diff
@@ -0,0 +1,134 @@
...
+            ->setDescription("Checks if your host's environment meets Manala's requirements (Vagrant, Ansible, etc.)")
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->addFormatter($output);
```
ogizanagi: `configureFormatter($output->getFormatter())` ?

src/Config/Requirement/Violation/RequirementViolationList.php
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function offsetUnset($offset)
+    {
+        unset($this->violations[$offset]);
+    }
+
+
```
ogizanagi: Extra blank lines

src/Config/Requirement/Violation/RequirementViolationList.php
```diff
@@ -0,0 +1,176 @@
...
+namespace Manala\Config\Requirement\Violation;
+
+use Manala\Config\Requirement\Common\RequirementLevel;
+
+/**
+ * Iterable collection of requirement violations.
+ *
+ * @author Xavier Roldo <xavier.roldo@elao.com>
+ */
+class RequirementViolationList implements \Iterator, \ArrayAccess, \Countable
```
ogizanagi: What do you think about simply inheriting from `\ArrayObject` ? IMHO this class is unlikely to need extending anything else later.

src/Command/CheckRequirements.php
```diff
@@ -0,0 +1,134 @@
...
+     * @return Requirement[]
+     */
+    private function getRequirements()
+    {
+        return [
+            new Requirement(
+                'vagrant',
+                Requirement::TYPE_BINARY,
+                RequirementLevel::REQUIRED,
+                '>= 1.8.0 < 1.8.5 || ^1.8.6', // /!\ Exclude vagrant 1.8.5 (Manala incompatible)
```
nervo: Maybe 1.8.4 || ^1.8.6 ?
I don't think we should support  1.8.0 -> 1.8.3, if i remember well, some versions are also buggy

src/Command/CheckRequirements.php
```diff
@@ -0,0 +1,134 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Command;
+
+use Manala\Config\Requirement\Common\RequirementLevel;
+use Manala\Config\Requirement\Factory\HandlerFactoryResolver;
+use Manala\Config\Requirement\Requirement;
+use Manala\Config\Requirement\RequirementChecker;
+use Manala\Config\Requirement\Violation\RequirementViolation;
```
nervo, line 17: Am i the only one prefering to have a single "Manala\Config\Requirement\Violation" use, and some "Violation\RequirementViolation" or "Violation\RequirementViolationLabelBuilder" inside code ?

src/Config/Requirement/Common/RequirementLevel.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * (c) Manala <contact@manala.io>
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Config\Requirement\Common;
+
+class RequirementLevel
```
ogizanagi: Usually declared `final`. :)

src/Command/CheckRequirements.php
nervo: 1.0.0 here :)
See https://github.com/Elao/symfony-standard/blob/master/README.app.md#requirements
And maybe, we should use the last relase, aka. 1.1.2
See https://github.com/vagrant-landrush/landrush/releases

src/Config/Requirement/Exception/MissingRequirementException.php
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Config\Requirement\Exception;
+
+class MissingRequirementException extends \Exception
+{
+
```
ogizanagi: extra blank line

src/Command/CheckRequirements.php
nervo: Same as landrush, we should use the last available release, aka. 2.1.1
See https://github.com/ansible/ansible/releases

src/Config/Requirement/Processor/AbstractProcessor.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $command = $this->getCommand($name);
+        $process = new Process($command);
+        $process->run();
+
+        if (!$process->isSuccessful()) {
+            throw new MissingRequirementException();
+        }
+
+        return $process->getOutput();
```
nervo, line 41: Hm, be careful that some binaries can use stderr to print some useful informations.

src/Config/Requirement/Factory/VagrantPluginHandlerFactory.php
```diff
@@ -0,0 +1,43 @@
...
+ * Factory that instantiates the concrete processor and version parser to handle vagrant plugin requirements.
+ *
+ * @author Xavier Roldo <xavier.roldo@elao.com>
+ */
+class VagrantPluginHandlerFactory implements HandlerFactoryInterface
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @return VagrantPluginProcessor
```
ogizanagi: I think it is not necessary to indicate the returned type here, below, and in the other factory.

src/Config/Requirement/Processor/VagrantPluginProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process($name)
+    {
+        $output = parent::process($name);
+
+        // If output is empty, it means that the vagrant plugin identified by $name is not installed:
+        if ($output == '') {
```
ogizanagi: If you do not use strict comparison, then why not using `empty` for semantic.

src/Config/Requirement/Processor/AbstractProcessor.php
```diff
@@ -0,0 +1,52 @@
...
+     * @return string|null Returns the standard output if tool is available, false otherwise.
+     */
+    public function process($name)
+    {
+        $command = $this->getCommand($name);
+        $process = new Process($command);
+        $process->run();
+
+        if (!$process->isSuccessful()) {
+            throw new MissingRequirementException();
```
ogizanagi: I'm not sure about this. Actually if the process is not successful, isn't it because of a logic issue in the code ? (bad command)

chalasr:
> we should use the last available release, aka. 2.1.1

@nervo I was more thinking that we should warn the user only if its system doesn't fit the minimal requirements (being `2.0.0` here and `0.18` for landrush, right?), then we could eventually advice to install the last stable release (which will not stay 2.1.1 for a long time, isn't it? :)). What do you think?

tests/Config/Requirement/RequirementCheckerTest.php
```diff
@@ -0,0 +1,162 @@
...
+            'vagrant',
+            Requirement::TYPE_BINARY,
+            RequirementLevel::REQUIRED,
+            '^1.8.2'
+        );
+
+        $this->processor->process('vagrant')->willReturn('vagrant 1.7.2');
+        $this->handlerFactory->getVersionParser()->willReturn(new BinaryVersionParser());
+        $this->requirementChecker->check($requirement, $violationList);
+        $this->assertEquals(1, count($violationList));
```
ogizanagi: Detail, but you can use `assertCount` here FYI (same for other occurrences) :)

tests/Config/Requirement/RequirementCheckerTest.php
```diff
@@ -0,0 +1,162 @@
...
+        $this->requirementChecker = new RequirementChecker(
+            $handlerFactoryResolver->reveal(),
+            new RequirementViolationLabelBuilder()
+        );
+    }
+
+    public function testCheckerShouldCreateAViolationIfRequiredBinaryIsMissing()
+    {
+        $violationList = new RequirementViolationList();
+        $this->assertEquals(0, count($violationList));
```
ogizanagi: `assertEmpty`

ogizanagi (approved):
It looks really great 😉

src/Config/Requirement/Processor/AbstractProcessor.php
ogizanagi: @nervo : Even if that's true, it seems very unlikely to help in order to find the version number, right ?
stderr should only be here for error, debug and things like progress bars and misc infos about the command execution. I doubt a command we'll execute in order to get the version number will print it to stderr when it's the meaningful information.

src/Command/CheckRequirements.php
ogizanagi: Unless there is a conflict between layers or names, I don't like much aliasing a namespace instead of the FQCN. :/

src/Config/Requirement/Processor/AbstractProcessor.php
nervo: I recently met this situation with nginx :)

nervo:
@chalasr see https://github.com/Elao/symfony-standard/blob/master/README.app.md#requirements
landrush is 1.0.0+
Let's discuss this point irl tomorow :)

chalasr:
@ogizanagi

> I'm not sure about this. Actually if the process is not successful, isn't it because of a logic issue in the code ? (bad command)

My guess is that @xavier-rdo deduces that a binary is not available through the fact the process fails. 
If I'm right, maybe we can find a more accurate way to do it, eventually taking some inspiration from the symfony [`PhpExecutableFinder`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Process/PhpExecutableFinder.php)

src/Config/Requirement/Processor/AbstractProcessor.php
ogizanagi: Erf. Then a solution might be to provide and call a `abstract extract(Output)/doProcess(Process $process)`— instead of overriding this method in children — in order to extract the version from the process instance.

ogizanagi:
(Answering to https://github.com/manala/manala/pull/19#issuecomment-249979978)

@chalasr : Right, but this should not be inside the abstract class I guess. (It's ok for the `BinaryProcessor`, but makes no sense for the `VagrantPluginProcessor` for instance).
Eventually this check might be done inside the abstract method mentioned [here](https://github.com/manala/manala/pull/19#discussion_r80780593)

chalasr (changes requested):

src/Command/CheckRequirements.php
chalasr: (Versions to be rediscussed with @nervo)

xavier-rdo:
Thanx to all for your feedbacks. 

Followings to your feedbacks are in the third commit ("Update/Improvements etc.")

Robin said : "_I was more thinking that we should warn the user only if its system doesn't fit the minimal requirements (being 2.0.0 here and 0.18 for landrush, right?), then we could eventually advice to install the last stable release (which will not stay 2.1.1 for a long time, isn't it?_"

@chalasr : for the moment, i don't implement recommendations about version, they are "required requirements" (they must be installed on the host with the expected version(s)) ; required executable but recommended version is not implemented yet (& seems to make things a little more complicated).

src/Config/Requirement/Processor/AbstractProcessor.php
```diff
@@ -49,4 +49,20 @@ public function process($name)
...
+    /**
+     * Override this method in concrete processor implementations if you want to grant a last chance to obtain
+     * information about the executable from the error output.
+     *
+     * Default hereby implementation treats any process failure as a missing executable and therefore throws a
+     * MissingRequirementException.
+     *
+     * @param string $errorOutput
+     *
+     * @throws MissingRequirementException
```
ogizanagi, line 62: Missing `@return` information ?

src/Command/CheckRequirements.php
chalasr: 👍

xavier-rdo: Done (Introduced a basic RequirementRepository with in-memory Requirements)

chalasr (approved):

src/Config/Requirement/RequirementRepository.php
```diff
@@ -0,0 +1,47 @@
...
+namespace Manala\Config\Requirement;
+
+use Manala\Config\Requirement\Common\RequirementLevel;
+
+class RequirementRepository
+{
+    /**
+     * @return Requirement[]
+     */
+    public function getRequirements()
```
chalasr: This method could be static, avoiding the need of the `$repository` variable [here](https://github.com/manala/manala/pull/19/files#diff-140f323652e32bc7d859c64c1cbf8eefR57)

src/Config/Requirement/Violation/RequirementViolationList.php
chalasr: 👍  for extending `\ArrayObject`  as only `offsetSet()` is specific to this implementation

src/Config/Requirement/Violation/RequirementViolationLabelBuilder.php
```diff
@@ -0,0 +1,51 @@
...
+                $isRequired ?
+                    '%s is missing. Please install it before proceeding.' :
+                    'You should consider installing %s.',
+                $requirement->getName()
+            );
+        } else {
+            $label .= sprintf(
+                $isRequired ?
+                    "Your %s version %s doesn't allow you to use this. Required version is %s." :
+                    'Your %s current version is %s. We recommend you to upgrade to version %s',
```
chalasr: missing ending dot here.

src/Config/Requirement/Violation/RequirementViolationLabelBuilder.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public function buildViolationLabel(Requirement $requirement, $currentVersion = null)
+    {
+        $label = sprintf('[%s] ', strtoupper($requirement->getLevelLabel()));
+        $isRequired = $requirement->isRequired();
+
+        if ($currentVersion === null) {
+            $label .= sprintf(
+                $isRequired ?
+                    '%s is missing. Please install it before proceeding.' :
```
chalasr, line 34: A break-line at the end of `$label` would make the help easier to read I think, at least there should be a space as we concatenate the help later.

src/Config/Requirement/Common/RequirementLevelHolderTrait.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * @return string
+     */
+    public function getLevelLabel()
+    {
+        switch ($this->level) {
+            case RequirementLevel::RECOMMENDED:
+                return 'Recommended';
+            case RequirementLevel::REQUIRED:
+
```
chalasr: No need for this blank line

src/Command/CheckRequirements.php
```diff
@@ -0,0 +1,103 @@
...
+     * @return string Formatted violation description displayed to user
+     */
+    private function getFormattedViolationDescription(RequirementViolation $violation)
+    {
+        $resultPattern = $violation->isRequired() ? '<error>%s</error>' : '<warning>%s</warning>';
+        $displayedText = $violation->getLabel();
+        $displayedText .= $violation->getHelp() ? ' '.$violation->getHelp() : '';
+
+        return sprintf($resultPattern, $displayedText);
+    }
```
chalasr, line 101: I would like to propose you [a small patch](https://github.com/manala/manala/commit/fb1d1e27d26593dfcd4c323c9e4675e7764a7ba8) which makes use of `SymfonyStyle` as output. I find it nice and we use it in other commands, what do you think?

src/Config/Requirement/Processor/VagrantPluginProcessor.php
chalasr: No strong opinion on this, let's keep this as is and keep the problem in mind (maybe open an issue for discussing about the whole windows support)

chalasr (changes requested):
Again, very nice job. I left some minor comments, then 👍

src/Config/Requirement/Violation/RequirementViolationList.php
xavier-rdo: Eh eh, i didn't really understand the \ArrayObject SPL class but i think i will use it intensively in the future (your recommendations and warnings about its usage are welcome ! ogizanagi already mentioned the limitations related to inheritance).

src/Config/Requirement/Processor/VagrantPluginProcessor.php
xavier-rdo: I'd rather try to stay windows-compatible as long as it is not painful. And for the current case, making the processor/versionParser for vagrant plugins windows compliant is not painful at all (but less 'elegant'). I will update the processor & versionParser accordingly.

@nervo: any opinion about that ?

tests/Config/Requirement/RequirementCheckerTest.php
xavier-rdo: Thanx Ogizanagi

src/Command/CheckRequirements.php
chalasr: As you said, I'll open a PR targeting this one

src/Config/Requirement/Processor/VagrantPluginProcessor.php
ogizanagi: Windows users developpers: 
Kill it with fire !

src/Config/Requirement/Processor/VagrantPluginProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process($name)
+    {
+        $output = parent::process($name);
+
+        // If output does not contain '$name', it means that the vagrant plugin is not installed:
+        if (!preg_match($name, $output)) {
```
chalasr: > PHP Warning:  preg_match(): Delimiter must not be alphanumeric or backslash in /srv/manala/manala/src/Config/Requirement/Processor/VagrantPluginProcessor.php on line 26

What about `if (false === strpos($output, $name))`

xavier-rdo: Merci Robin, en plus, c'est écrit noir sur blanc dans la doc (privilégier strpos) :

```
Astuce
N'utilisez pas preg_match() si vous voulez uniquement savoir si une chaîne est contenue dans une autre. Utilisez dans ce cas les fonctions strpos() ou strstr(), qui sont beaucoup plus rapides.
```

chalasr (approved):

chalasr:
Thank you @xavier-rdo

PR #22, opened 2016-09-25
Refactor structure for next needs

chalasr:


src/Env/Config/Manala.php
```diff
@@ -0,0 +1,28 @@
...
+ * (c) Manala <contact@manala.io>
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Env\Config;
+
+/**
+ * Make.
```
ogizanagi: Make ? Make what ? 😆 

EDIT: oh, just saw it was simply a copy/paste from above 😅 . Anyway may be removed :)

chalasr: @ogizanagi: The biggest troll ever 😆 

It corresponds to the unix`make` utility, same as `Vagrant` for `vagrant` and `Ansible` for `ansible`. But you're right it's misleading, I'll expand this description

ogizanagi: @chalasr : No no no. I mean... unless `make` has something to do with the `Manala` class on which this docblock is... :trollface:

ogizanagi: But appart from this, I'm not fan of such docblocks simply repeating the class name without bringing more information, indeed. :)

chalasr: lol, sorry about the confusion :trollface:

PR #23, opened 2016-09-27
Remove composer.lock from .gitignore

xavier-rdo:
Is there an explicit reason to have the _composer.lock_ file in _.gitignore_ and therefore unversioned ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 /.build/
 /vendor/
-/composer.lock
 /manala.phar
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,1921 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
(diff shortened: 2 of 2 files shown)


chalasr:
@xavier-rdo In general, the need of having dependencies locked is less important for third party tools. Here, we only use some symfony components which have a very structured release process and a real [backward compatibility promise](http://symfony.com/doc/current/contributing/code/bc.html). So, the version constraint present in the composer.json should be enough to guarantee stability, no need for locking them to a specific patch, as we are sure any next point release will not break anything, or at least if it does, we are sure a fix will be released in good delays.

If a sensible dependency comes to be added (concerning security for instance), then we could stop ignoring it. For now, I feel safe with the current requirements.

chalasr:
Also we already discussed about this in https://github.com/manala/manala/pull/12#pullrequestreview-618923

PR #24, opened 2016-09-27
Add php-cs-fixer check to Travis

ogizanagi:
(only executed on the PHP 7 build for speed reasons. Also, `hhvm` conflicts with it)

Enabling the cache on travis requires to move the cache file in a dedicated repository we can cache (only folders can be cached on Travis). But IMHO, it's not worth it for so few classes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,17 @@
 language: php
 
-php:
-    - 5.6
-    - 7.0
-    - hhvm
+env:
+    global:
+        - CHECK_PHP_SYNTAX="no"
+        - PATH="$PATH:$HOME/.composer/vendor/bin"
 
 matrix:
     fast_finish: true
+    include:
+        - php: 7.0
+          env: CHECK_PHP_SYNTAX="yes"
+        - php: 5.6
+        - php: hhvm
     allow_failures:
         - php: hhvm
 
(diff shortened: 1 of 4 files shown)


chalasr:
Thank you @ogizanagi 🍺

chalasr:
(just discovered the new [rebase and merge](https://github.com/blog/2243-rebase-and-merge-pull-requests) github feature ❤️ )

ogizanagi:
Yes, that's a good thing for co-working on branches IMHO. But I'd rather keep merge commits on `master`, don't you ?

ogizanagi:
Woops. Just saw you required-dev `friendsofphp/php-cs-fixer` in the `composer.json`.
Should we remove the global install from Travis then?

chalasr:
> But I'd rather keep merge commits on master 

I do not care much about having merge commits since pretty much all changes are merged from pull requests, it makes easy for me to differentiate real changes from merges between static branches (not PR related ones). However I'd be glad to have your thoughts about why we should keep them, if any :)

> Woops

Woops..., missed it. I'll quickly remove the global install from the travis file if you don't do it before (not forced to open a PR for such)

ogizanagi:
The PR was ready 😉 

> I do not care much about having merge commits since pretty much all changes are merged from pull requests, it makes easy for me to differentiate real changes from merges between static branches (not PR related ones). However I'd be glad to have your thoughts about why we should keep them, if any :)

Simply pleased by the graph with the merge commit on the `master`. 👅 I like the workflow.
Most PRs are squashed in a single commit, but some may require legitimate atomic commits. The merge commit and the graph visually help as a reminder in such situations (outside from github ability to show the related PR in the commit view).

But indeed that's not essential.

PR #26, opened 2016-09-28
[Travis] Remove composer self update from Travis

ogizanagi:
I made the PR FYI: it seems to be already done by Travis: https://travis-ci.org/manala/manala/jobs/163382901#L147 ^^'


chalasr:
Good catch @ogizanagi! Thanks

PR #29, opened 2016-09-30
Allow configuring dependencies in setup

chalasr:
- [x] In a `packages.yaml` (better name?), define the available packages per env type (only symfony atm) with their version bound and default version
- [x] Ask the user which packages should be installed and (if chosen) in which version
- [x] Dump `ansible/group/vars/app.yml` after merging the `app_options` key with the configured packages 
- [x] Update tests

### DIFF
--- a/src/Command/Setup.php
+++ b/src/Command/Setup.php
@@ -12,9 +12,14 @@
 namespace Manala\Command;
 
 use Manala\Env\Config\Variable\AppVendor;
+use Manala\Env\Config\Variable\Dependency\Dependency;
+use Manala\Env\Config\Variable\Dependency\VersionBounded;
+use Manala\Env\Config\Variable\MakeTarget;
 use Manala\Env\Dumper;
 use Manala\Env\EnvEnum;
 use Manala\Env\EnvFactory;
+use Manala\Env\Metadata\MetadataBag;
+use Manala\Env\Metadata\MetadataParser;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
@@ -57,18 +62,53 @@ class Setup extends Command
         $io->setDecorated(true);
         $io->comment(sprintf('Composing your <info>%s</info> environment', (string) $envType));
 
-        $vars = new AppVendor(
+        $appVendor = new AppVendor(
             $io->ask('Vendor name', null, [AppVendor::class, 'validate']),
             $io->ask('App name', null, [AppVendor::class, 'validate'])
         );
-        $env = EnvFactory::createEnv($envType);
 
-        foreach (Dumper::dump($env, $vars, $cwd) as $dumpTarget) {
+        $envMetadata = MetadataParser::parse($envType);
+        $postProvisionTask = new MakeTarget('install', $envMetadata->get('script.post_provision'));
+        $env = EnvFactory::createEnv($envType, $appVendor, $postProvisionTask, $this->setupDependencies($io, $envMetadata));
+
+        foreach (Dumper::dump($env, $appVendor, $cwd) as $dumpTarget) {
             $io->writeln(sprintf('- %s', str_replace($cwd.'/', '', $dumpTarget)));
         }
 
         $io->success('Environment successfully configured');
 
         return 0;
+    }
+
+    private function setupDependencies(SymfonyStyle $io, MetadataBag $metadata)
+    {
+        foreach ($metadata->get('packages') as $name => $settings) {
+            $defaultEnabled = $settings['enabled'] ?: false;
+            $defaultVersion = isset($settings['default']) ? $settings['default'] : null;
+            $enabled = $settings['required'] ?: $io->confirm(sprintf('Install %s?', $name), $defaultEnabled);
+
+            if (null === $defaultVersion) {
+                yield new Dependency($name, $enabled);
+
+                continue;
+            }
+
+            if (false === $enabled) {
+                yield new VersionBounded($name, $enabled, $defaultVersion);
+
+                continue;
+            }
+
+            $versionConstraint = $settings['constraint'];
+            $requiredVersion = $io->ask(
+                sprintf('%s version? (%s)', $name, $versionConstraint),
+                $defaultVersion,
+                function ($version) use ($versionConstraint) {
+                    return VersionBounded::validate($version, $versionConstraint);
+                }
+            );
+
+            yield new VersionBounded($name, $enabled, $requiredVersion);
+        }
     }
 }
--- a/src/Env/Config/Config.php
+++ b/src/Env/Config/Config.php
@@ -11,6 +11,7 @@
 
 namespace Manala\Env\Config;
 
+use Manala\Env\Config\Variable\Variable;
 use Manala\Env\EnvEnum;
 
 /**
@@ -26,11 +27,17 @@ abstract class Config
     protected $envType;
 
     /**
+     * @var Variable[]
+     */
+    protected $vars;
+
+    /**
      * @param EnvEnum $env
      */
-    public function __construct(EnvEnum $envType)
+    public function __construct(EnvEnum $envType, Variable ...$vars)
     {
         $this->envType = $envType;
+        $this->vars = $vars;
     }
 
     /**
@@ -52,18 +59,15 @@ abstract class Config
      */
     public function getOrigin()
     {
-        return MANALA_DIR.'/src/Resources/'.$this->envType.'/'.$this->getPath();
+        return new \SplFileInfo(MANALA_DIR.'/src/Resources/'.$this->envType.'/'.$this->getPath());
     }
 
     /**
      * Returns the template to render (or its path).
      *
-     * @return \SplFileInfo|string
+     * @return \SplFileInfo|null
      */
-    public function getTemplate()
-    {
-        return $this->getOrigin();
-    }
+    abstract public function getTemplate();
 
     /**
      * Returns the path name of the configuration file or directory.
@@ -73,4 +77,14 @@ abstract class Config
      * @return string
      */
     abstract public function getPath();
+
+    /**
+     * Returns the variables to be used for rendering the template.
+     *
+     * @return array|null
+     */
+    public function getVars()
+    {
+        return $this->vars;
+    }
 }
--- a/src/Env/EnvFactory.php
+++ b/src/Env/EnvFactory.php
@@ -14,6 +14,8 @@ namespace Manala\Env;
 use Manala\Env\Config\Ansible;
 use Manala\Env\Config\Make;
 use Manala\Env\Config\Vagrant;
+use Manala\Env\Config\Variable\AppVendor;
+use Manala\Env\Config\Variable\MakeTarget;
 
 /**
  * Provides Env instances.
@@ -27,8 +29,12 @@ class EnvFactory
      *
      * @return Env
      */
-    public static function createEnv(EnvEnum $type)
+    public static function createEnv(EnvEnum $type, AppVendor $appVendor, MakeTarget $postProvisionTask, \Traversable $dependencies)
     {
-        return new Env(new Ansible($type), new Vagrant($type), new Make($type));
+        return new Env(
+            new Ansible($type, ...$dependencies),
+            new Vagrant($type, $appVendor),
+            new Make($type, $postProvisionTask)
+        );
     }
 }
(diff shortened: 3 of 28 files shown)


src/Env/Config/Config.php
```diff
@@ -42,28 +50,25 @@ public function getFiles()
     {
         $origin = $this->getOrigin();
 
-        yield ($origin instanceof \SplFileInfo) ? $origin : new \SplFileInfo($origin);
+        yield ($origin instanceof SplFileInfo) ? $origin : new SplFileInfo($origin);
```
ogizanagi: Detail, but why those changes ? i'm not fan of aliasing things from the global namespace.

src/Env/EnvFactory.php
```diff
@@ -27,8 +29,8 @@ class EnvFactory
      *
      * @return Env
      */
-    public static function createEnv(EnvEnum $type)
+    public static function createEnv(EnvEnum $type, AppVendor $appVendor, Dependency ...$dependencies)
```
ogizanagi: new arguments missing from phpdoc

ogizanagi: BTW, not sure I like the varargs in this factory method. When you'll introduce a new argument, this one'll have to kept being the last one. 
If the newly introduced var is an array too, what would be the legitimate reason to have one variadic but another not ?

IMHO variadic functions/methods must not be overused and does only make sense to very specific cases.

ogizanagi:
> In a `packages.yaml` (better name?), define the available packages per env type (only symfony atm) with their version bound and default version

`packages.yaml` looks good to me, but if we need to introduce something else later, instead of creating another file, WDYT about a `manala.yml` file with the following content:

``` yaml
setup:
    packages:
        php:
            enabled:    true
            default:    '7.0'
            constraint: '5.4|5.5|5.6|7.0'
        mysql:
            enabled:    true
            default:    '5.6'
            constraint: '5.6|5.7'
        postgresql:
            enabled:    false
            default:    '9.5'
            constraint: '9.4|9.5'
        mongodb:
            enabled:    false
            default:    '3.2'
            constraint: '3.0|3.2'
        elasticsearch:
            enabled:    false
            default:    '1.7'
            constraint: '1.5|1.6|1.7|2'
        nodejs:
            enabled:    false
            default:    '6'
            constraint: '0.10|0.12|4|5|6'
        redis:
            enabled:    false

# ========== Related to https://github.com/manala/manala/pull/30#discussion_r81446586

build:
    install: 
        - composer install --verbose
        - bin/console doctrine:database:create --if-not-exists
```

(or without the `setup` key if you do think there will only exist things for the `Setup` command in here).

src/Env/Config/Config.php
chalasr: Simply because it is used many times in the class but I'll remove the use, after all there should not have any gain in doing that.

src/Env/EnvFactory.php
chalasr: I agree with you, the signature may change and type hints should be consistent, I'll change to `array $dependencies` so we don't have to do it later.

However, Imho it makes sense to use variadics here, it should not have any particular cost to use them so even if they indeed solve some specific needs, I don't think it's wrong to use them for basic needs (i.e. do not have to iterate over all elements of an array for being sure they are of the good type).

ogizanagi: > i.e. do not have to iterate over all elements of an array for being sure they are of the good type

Sure. But that's a limitation of PHP to not be able to valid array items (generics may help regarding this in PHP 7.2 or later). 
But variadics aim is not about validation, and thus are misused if you only use variadics for this reason.

chalasr:
I like your idea! 

My only doubt is about the filename, it is not excluded that next needs involve to create a `manala.yml` inside of the consumer's project (at the root dir or not, as a dotfile or not), which would allow running the setup command without interaction, the dependencies being configured into with shared directories, app/vendor and such.., for which the content would be validated using our metadata files (which contains constraints and defaults).

Maybe `metadata.yml`, `env.yml`, `config.yml`? (please, help 😅)

Also I think to make use of the `symfony/config` component for using the `imports` feature (I'm pretty sure some envs will share the same `install` and some common `packages`).

ogizanagi:
> My only doubt is about the filename, it is not excluded that next needs involve to create a manala.yml inside of the consumer's project (at the root dir or not, as a dotfile or not), which would allow running the setup command without interaction, the dependencies being configured into with shared directories, app/vendor and such.., for which the content would be validated using our metadata files (which contains constraints and defaults).

First:

This file will live in the root of the project to manalize, right ? Thus it should be created by the user before calling the executable to manalize the project IIUC ? So it doesn't prevent us to have a `src/Resources/symfony/manala.yml` file ? If we do not copy this file at the root of the project, it won't collide , right ? 

Then, I've got many concerns about this (but maybe I've already mentioned it):

To me, the manala executable is a one time use per project, in order to init it. I don't see much value to this process, given that the result of the "manalization" will be commited very soon in the repository. Then, we'll directly use `./ansible/*` files in order to tweak the manala configuration during the project development lifecycle.

chalasr:
You're right about it will not collide, 👍  for `manala.yml`.

I know your thoughts about the usefulness of the whole `build` command+process, but I see it like a good opportunity to better control the provisioning than we can do it using the `make setup` command, especially for giving better error outputs so make fails easier to debug (not good atm, it needs some work) and better managing them than vagrant does it e.g. by destroying the vm or retrying depending on the fail. It will need some work to make it useful.
Also, it will be nice to be able to test our build process on different environments, I'm going to add a `php 5.6` version of the `BuildTest`.
Of course you can always bypass it using the `make` task directly, I do not have any problem with that.

About the `manala.yml`, I'm quietly changing my mind on it.
It's indeed not needed to make this file part of the git repository index, as most of the time people will run the `setup` command interactively once, run the `build` one and move on. 
However, this file may be useful to have as a datastore, placed into the `ansible/` directory, persisting the configuration made in the `setup` command, then used by the `build` (and maybe by the `deploy` in the future) to ensure the working directory is ready (checking that storage directories are empty otherwise the provisioning fails at creating symlinks, verifying the presence of some env-specific files and such use cases that I may not be aware of atm). 

Your opinion (and all others) are really important, also this point will need to be seriously discussed at the good moment

chalasr:
~~Hmm, depending on the last part of my previous comment, `manala.yml` may collide ^^~~ this file will never go on the consumer's project.

ogizanagi:
> However, this file may be useful to have as a datastore, placed into the ansible/ directory, persisting the configuration made in the setup command, then used by the build (and maybe by the deploy in the future) to ensure the working directory is ready (checking that storage directories are empty otherwise the provisioning fails at creating symlinks, verifying the presence of some env-specific files and such use cases that I may not be aware of atm).

To me, it looks like it's the role of manala deploy ansible role, which will evolves along the project lifecycle.
The manala command has already done its job.

chalasr:
Agreed. We are dumping all the settings done in the `setup` command to the ansible config files so we can refer to them for doing verifications and performing actions depending on. The only interest of generating a config file stay to reuse it, and I agree there is no real need for at the moment, and maybe there will never. 

About the shared directories and such, we need to keep @nervo informed, as you pointed out, for making ansible roles evolve as long as we discover new needs, especially for being able to handle existing projects, which is the first motivation for this tool.

It's indeed not the role of this to prevent the behavior of the ansible roles, we should adapt them when it make sense 👍

src/Env/Metadata/MetadataBag.php
```diff
@@ -0,0 +1,124 @@
...
+     * @param string|null $fullPath
+     *
+     * @return \LogicException
+     */
+    private function didYouMean($search, array $possibleMatches, $fullPath)
+    {
+        $minScore = INF;
+
+        foreach ($possibleMatches as $try) {
+            $distance = levenshtein($search, $try);
```
chalasr, line 101: @xavier-rdo here is a use of levenstein :)

chalasr:
@ogizanagi I made changes :)

tests/Env/Config/RendererTest.php
```diff
@@ -0,0 +1,70 @@
...
+        $var = $this->prophesize(Variable::class);
+        $var->getReplaces()->willReturn(['{{ placeholder }}' => 'bar']);
+        $config = $this->prophesize(Config::class);
+        $config->getVars()->willReturn([$var->reveal()]);
+        $config->getTemplate()->willReturn(new \SplFileInfo(self::$tempdir.'/template'));
+
+        $this->assertSame('foo bar baz', Renderer::render($config->reveal()));
+    }
+
+    public function testRenderYamlWithReplacesAtDeep()
```
ogizanagi: What is this test case for ? 😕

src/Env/Config/Variable/MakeTask.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+namespace Manala\Env\Config\Variable;
+
+/**
+ * Manala "app" and "vendor" environment variables.
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+final class MakeTask implements Variable
```
ogizanagi: I find `MakeTask` confusing (I though it was something that Manala will execute at some point instead of a `Variable`). 
What do you think about `MakeTarget` ?

src/Env/Config/Variable/MakeTask.php
```diff
@@ -0,0 +1,56 @@
...
+ * (c) Manala <contact@manala.io>
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Env\Config\Variable;
+
+/**
+ * Manala "app" and "vendor" environment variables.
```
ogizanagi: To remove or update 😉

src/Command/Setup.php
```diff
@@ -57,18 +62,53 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+    private function setupDependencies(SymfonyStyle $io, MetadataBag $metadata)
+    {
+        foreach ($metadata->get('packages') as $name => $settings) {
+            $defaultEnabled = $settings['enabled'] ?: false;
+            $defaultVersion = isset($settings['default']) ? $settings['default'] : null;
+            $enabled = $settings['required'] ?: $io->confirm(sprintf('Install %s?', $name), $defaultEnabled);
+
+            if (null === $defaultVersion) {
+                yield new Dependency($name, $enabled);
```
ogizanagi, line 91: Can you explain why using a generator in this context and how it is useful ? :)

ogizanagi (approved):

ogizanagi:
👍

tests/Env/Config/RendererTest.php
chalasr: Oops.. forget to finish it. It is for testing that dependencies are properly filled under the `app_options` level of the `group_vars/app.yml` template

tests/Env/Config/RendererTest.php
```diff
@@ -0,0 +1,70 @@
...
+        $var = $this->prophesize(Variable::class);
+        $var->getReplaces()->willReturn(['{{ placeholder }}' => 'bar']);
+        $config = $this->prophesize(Config::class);
+        $config->getVars()->willReturn([$var->reveal()]);
+        $config->getTemplate()->willReturn(new \SplFileInfo(self::$tempdir.'/template'));
+
+        $this->assertSame('foo bar baz', Renderer::render($config->reveal()));
+    }
+
+    public function testRenderYaml()
```
chalasr: Same for this one, there isn't any yaml here

src/Env/Config/Variable/MakeTask.php
chalasr: 👍

tests/Env/Config/RendererTest.php
ogizanagi: According to Yaml lint, `foo bar baz` is valid Yaml, though :trollface:

src/Command/Setup.php
chalasr: Using a generator here avoids building an array internally (of an unknown number of elements that can be potentially big)

chalasr: You're right that there is no gain in term of memory at using a generator here. However there is no cost and I find nicer to deal with an Iterator rather than an array here :)

tests/Env/Config/RendererTest.php
chalasr: Fixed

chalasr:
Tests updated, I think this is ready

ogizanagi:
Now it's merged. Congrats @chalasr 👏

PR #30, opened 2016-10-01
Avoid using make install in build

chalasr:


src/Handler/Build.php
```diff
@@ -44,7 +44,7 @@ public function __construct($cwd)
         $vagrantTasks = [
             'up --no-provision',
             'provision',
-            'ssh -- "cd /srv/app && make install"',
+            'ssh -- "cd /srv/app && composer install --verbose && bin/console doctrine:database:create --if-not-exists"',
```
ogizanagi, line 47: Shouldn't this depend on the env that was setup ?

chalasr: You're right, maybe calling the makefile stay the simpler way... as it is created per env.
If you have some suggestion... I take :)

ogizanagi: Looks like something that could be configured per env in a [`manala.yml` file](https://github.com/manala/manala/pull/29#issuecomment-250901383) under a `build.install` key ? 😄

chalasr: Really nice input!

chalasr:
Fixed in #29

PR #31, opened 2016-10-01
Make use of SymfonyStyle for displaying violations

xavier-rdo:
I've created the merge request so that we don't forget Robin's improvements. I will rebase in order to remove unwanted commits and resolve conflicts, next monday.

### DIFF
--- a/src/Command/CheckRequirements.php
+++ b/src/Command/CheckRequirements.php
@@ -22,6 +22,7 @@ use Symfony\Component\Console\Formatter\OutputFormatterInterface;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
 
 /**
  * Checks if the host's environment meets Manala's requirements (Vagrant, Ansible, etc.).
@@ -46,8 +47,7 @@ class CheckRequirements extends Command
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $this->configureFormatter($output->getFormatter());
-
+        $output = new SymfonyStyle($input, $output);
         $violationList = new RequirementViolationList();
         $requirementChecker = new RequirementChecker(
             new HandlerFactoryResolver(),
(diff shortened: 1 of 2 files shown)


ogizanagi:
@chalasr , @xavier-rdo : I've rebase the branch and appended my own suggestions:

The problem with the yellow background is that it is often unreadable with a white foreground:

screenshot 2016-10-01 a 10 04 21

That's the reason the `SymfonyStyle::warning()` method does not use the yellow background, because it has poor chances to have a sufficient contrast (of course it depends on your terminal ANSI colors config).

However, I do not share this conviction, and I'd like to keep it yellow. Simply changing the foreground for black (like it's done for `SymfonyStyle::success()` should be enough:

screenshot 2016-10-01 a 10 04 35

Finally, appart from minor tweaks in the code, I've prepended a new line to the command output, because `SymfonyStyle::write(ln)()` does not auto prepend automatically, as it's the base method used to write anything.
The little note at the end of the command was also changed to use `SymfonyStyle::note()`.

screenshot 2016-10-01 a 10 13 57

xavier-rdo:
Beautiful, ogizanagi ! 👍

chalasr:
Nice 👍

chalasr:
I have to make you aware of something if you aren't already: 

We are not running the whole test suite on travis because the build being ran in a docker container, we can't properly up a vagrant machine (trust me, I spent some nights to try all the things, a day we'll have a jenkins stack for running this important part of the suite).
So, when doing changes impacting sensible processes, please run `vendor/bin/phpunit` without the `--exclude-group infra` option (it may take some long minutes to finish) to ensure all is working before merging. 

Running it with the option should be enough most times, and it should be during the whole dev phase. Thank you!

ogizanagi:
Good advice 👍 
I saw in the `travis.yml` file how were tests launched, so I knew that, but didn't launched the suite here before merging ^^'. It does seem there is an issue with the current master, but not related to this PR however.

Anyway, don't you think we should append a section about tests in the `README.md` or a `CONTRIBUTING.md` file ?

chalasr:
@ogizanagi Note that this test depends on the good provisioning, which may break sometimes, that's also why it is (at the moment) not a bad thing that they're not part of our CI. Also please note that you'll be asked for root permissions.

👍  For documenting this, I already have a [wip](https://github.com/manala/manala/tree/usage#manala) for the usage but not for contributing.

PR #33, opened 2016-10-02
Remove build command

chalasr:
The first motivation for having this command was to have a better control over the VM provisioning,  a better error catching and a better behavior in case of fail, with btw a nice output.

However, it would be so hard to maintain for not much value.
So let's stay focused on the `setup` which is the main feature of this tool, lots of nice stuff can (and have to) be done on this part.


ogizanagi:
The `README.md` should be updated as well.

But let's wait for @nervo and @hyvs opinions about this. Maybe they have something else in mind that could live inside the `manala/manala` repository and bring enough value without being redondant with the manala roles config (and the `Makefile`).

nervo:
Ready to discuss tomorow morning :)

src/Exception/HandlingFailureException.php
```diff
@@ -1,16 +0,0 @@
...
- *
- * (c) Manala <contact@manala.io>
- *
- * For the full copyright and license information, please refer to the LICENSE
- * file that was distributed with this source code.
- */
-
-namespace Manala\Exception;
-
-class HandlingFailureException extends \RuntimeException
```
ogizanagi, line 14: This one is to keep (used by `Diff` and should probably be added to the `@throws` docblock section of [`Manala\Handler\Handler::handle()`](https://github.com/manala/manala/blob/master/src/Handler/Handler.php#L30))

ogizanagi (changes requested):

src/Exception/HandlingFailureException.php
chalasr: Good catch

chalasr: Fixed

ogizanagi (approved):

chalasr:
The naming has changed in the `manala.yml` mapping, now it contains two targets `setup` and `install`, both have placeholders in the makefile, replaced at manalizing.

The `BuildTest` (now calling `make setup`) is green.

src/Env/EnvFactory.php
```diff
@@ -24,16 +23,17 @@
...
-    public static function createEnv(
-        EnvEnum $type,
-        AppName $AppName,
-        MakeTarget $postProvisionTask,
-        \Traversable $dependencies
-    ) {
+    /**
+     * @param EnvEnum $type
+     *
+     * @return Env
```
ogizanagi: the whole docblock can be removed (was removed in #40 btw)

src/Resources/symfony/manala.yml
```diff
@@ -36,7 +36,14 @@ packages:
...
         enabled:    false
 
-script:
-    post_provision:
+make_targets:
+    install:
         - composer install --verbose
         - bin/console doctrine:database:create --if-not-exists
+    setup:
+        - if [ -d  "./var/cache" ]; then rm -rf ./var/cache && mkdir ./var/cache; fi;
```
chalasr, line 44: This is a workaround mandatory until @nervo automagically dumps all the files from these directories into a temp dir then redump them after provisioning (I have not the heart to write such in the makefile, this is already so ugly)

chalasr: Here is the error output:

> fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "gid": 20, "group": "dialout", "mode": "0755", "msg": "the directory /srv/app/var/logs is not empty, refusing to convert it", "owner": "501", "path": "/srv/app/var/logs", "size": 102, "state": “directory”

Because of a `.gitkeep` installed by the symfony standard edition and files created by composer scripts

ogizanagi (changes requested):
Don't forget to remove the usage in `README.md` :)

ogizanagi (approved):

src/Resources/symfony/manala.yml
```diff
@@ -36,7 +36,14 @@ packages:
         required:   false
         enabled:    false
 
-script:
-    post_provision:
+make_targets:
```
chalasr, line 39: Note that this allows to have only one generic makefile. Most of the targets that aren't in this yaml file are not specific to the env type at all.

PR #35, opened 2016-10-03
Ask for app name only instead of app+vendor

### LINKED ISSUE #28: The vendor name should not be required
I want to create a simple app without vendor name (to have a domain like app.dev and not app.vendor.dev)

chalasr:
- ~~Add `ManalaStyle` + `QuestionHelper` for bypassing the SymfonyQuestionHelper which doesn't allow empty inputs for `ask()` (see https://github.com/symfony/symfony/pull/20141)~~
- [x] Ask for App name instead of app and vendor separately (btw things become easier)

Closes #28


src/Command/Setup.php
```diff
@@ -55,11 +55,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $io->setDecorated(true);
-        $io->comment(sprintf('Composing your <info>%s</info> environment', (string) $envType));
+        $io->comment(sprintf('Start composing your <info>%s</info> environment', (string) $envType));
 
-        $vars = new AppVendor(
-            $io->ask('Vendor name', null, [AppVendor::class, 'validate']),
-            $io->ask('App name', null, [AppVendor::class, 'validate'])
+        $vars = new Host(
+            $io->ask('App name', 'app', [Host::class, 'validate']),
+            $io->ask('Vendor name (optional)', false, [Host::class, 'validate'])
```
chalasr: Using `false` as default value is a workaround. See https://github.com/symfony/symfony/pull/20141

ogizanagi:
Looks good to me 👍 :)

chalasr:
As discussed, I finally propose to ask for the hostname instead of the vendor+app name, this PR has been updated

src/Command/Setup.php
```diff
@@ -55,15 +55,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $io = new SymfonyStyle($input, $output);
         $io->setDecorated(true);
-        $io->comment(sprintf('Composing your <info>%s</info> environment', (string) $envType));
+        $io->comment(sprintf('Start composing your <info>%s</info> environment', (string) $envType));
 
-        $vars = new AppVendor(
-            $io->ask('Vendor name', null, [AppVendor::class, 'validate']),
-            $io->ask('App name', null, [AppVendor::class, 'validate'])
-        );
+        $host = $io->ask('Host name', 'app', [Host::class, 'validate']);
```
chalasr: I'm afraid that people may input `[host].com` instead of `[host]` for instance, but I can't find the good wording for mentioning that `.dev` will be automatically appended to this value so there's no need for doing it yourself, while staying concise of course :) Any clue?

chalasr:
@nervo It would be nice to have your opinion before merging this one

nervo:
Hmmm, 'looks like a little revolution :)
Btw, i'm not 100% convinced... :coffee: / :smoking: tomorrow morning ?

chalasr:
Sure!

src/Command/Setup.php
ogizanagi: Giving an example before asking the question is probably the best way to go :)

The default value already show that the top level domain is not necessary.

BTW, maybe we can set the default value to the current directory name instead of simply "app" ?

chalasr: Hope the default value did the job but still, if someone has something short enough to be wrapped into parenthesis and added at the end of the question label... I take!

> BTW, maybe we can set the default value to the current directory name instead of simply "app" ?

👍  Looks like a nice idea to me, but I fear it's too much at once for @nervo... :trollface:

chalasr:
- [x] Change Host in App

@nervo What do you think about https://github.com/manala/manala/pull/35#discussion_r81831817 proposal?

ogizanagi:
Needs rebase but looks ready to be merged to me 👍

PR #37, opened 2016-10-03
[WIP] Add manala diff command

ogizanagi:
This is a working draft for the `manala diff` command that will generate the diff with the targeted project and manala env templates and will be enhanced soon (mainly code design).

Usage:

``` sh
php manala.phar diff tests/fixtures/Command/DiffTest/AcmeProject/ --env=symfony
```

Patch can be applied by using PHPStorm ( ❤️ ) or by executing:

``` sh
 patch -p1 < my_patch.patch
```

inside the working directory.

PR #38, opened 2016-10-03
Add manala diff command

ogizanagi:
> https://github.com/manala/manala/pull/37 was improbably closed (missclick) in the exact same time I pushed force the branch. This prevents us from re-opening the previous PR...
### Description

The `manala diff` command generate the diff with the targeted project and manala env templates.

screenshot 2016-10-05 a 00 31 52
### Usage:

``` sh
# inside the manalized project with git (repo must be init):
manala diff
manala diff -q | git patch
```

screenshot 2016-10-04 a 22 57 28
#### Full:

``` sh
manala diff tests/fixtures/Command/DiffTest/AcmeProject/ --env=symfony
```

If saved in a file, patch can be applied by using PHPStorm ( ❤️ ) or by executing:

``` sh
 git patch my_patch.patch
```

inside the working directory.
### Known issues

File mode changes cannot be applied by the `patch` util. It means that using the next `manala diff` after applying the changes will still show file mode changes, and calling `patch -p1` again will return:

> patch: ***\* Only garbage was found in the patch input.

That should not be an issue by using `git apply`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: true
 language: php
 
 env:
@@ -19,6 +20,9 @@ cache:
         - $HOME/.composer/cache
 
 before_install:
+    - sudo add-apt-repository -y ppa:git-core/ppa
+    - sudo apt-get update && sudo apt-get install git-man && sudo apt-get install -y git
+    - git --version
     - phpenv config-rm xdebug.ini || true
 
 install: composer update --prefer-dist --no-interaction --optimize-autoloader
--- a/src/Command/Diff.php
+++ b/src/Command/Diff.php
@@ -1,0 +1,110 @@
+<?php
+
+/*
+ * This file is part of the Manala package.
+ *
+ * (c) Manala <contact@manala.io>
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Command;
+
+use Manala\Env\EnvEnum;
+use Manala\Exception\HandlingFailureException;
+use Manala\Handler\Diff as DiffHandler;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\ConsoleOutput;
+use Symfony\Component\Console\Output\NullOutput;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * @author Maxime STEINHAUSSER <maxime.steinhausser@gmail.com>
+ */
+class Diff extends Command
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('diff')
+            ->setDescription('Computes the diff between the current project and the Manala templates.')
+            ->addArgument('cwd', InputArgument::OPTIONAL, 'The path of the application', getcwd())
+            ->addOption('env', null, InputOption::VALUE_OPTIONAL, 'One of the supported environment types', 'symfony')
+            ->setHelp(<<<EOTXT
+{$this->getDescription()}
+
+The command output can be redirect to a file in order to create a patch to apply later, or for sharing:
+<info>
+manala diff > my_patch.patch 
+</>
+You can apply the generated patch by executing one of the following command in the project working directory:
+Using <comment>git</>:
+<info>
+$ manala diff | git apply
+$ git apply my_patch.patch
+</>
+Using <comment>patch</>:
+<info>
+$ manala diff | patch -p1
+$ patch -p1 < my_patch.patch
+</>
+As the <comment>git diff</> util, the command exit code is 1 if a diff is detected, 0 otherwise.
+EOTXT
+            )
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $cwd = realpath($input->getArgument('cwd'));
+        $envType = EnvEnum::create($input->getOption('env'));
+
+        if (!is_dir($cwd)) {
+            throw new \RuntimeException(sprintf('The working directory "%s" doesn\'t exist.', $cwd));
+        }
+
+        $handler = new DiffHandler($envType, $cwd, $output->isDecorated());
+
+        try {
+            $handler->handle(function ($type, $buffer) use ($output) {
+                $output->write($buffer);
+            });
+        } catch (HandlingFailureException $e) {
+            $io = new SymfonyStyle($input, $output);
+            $io->error(['An error occurred during the process execution:', $handler->getErrorOutput()]);
+
+            return $handler->getExitCode();
+        }
+
+        $errIo = $this->getErrIo($input, $output);
+
+        if (!$handler->hasDiff()) {
+            $errIo->success('No diff found.');
+        }
+
+        return $handler->getExitCode();
+    }
+
+    private function getErrIo(InputInterface $input, OutputInterface $output)
+    {
+        if (!$output instanceof ConsoleOutput) {
+            return new SymfonyStyle($input, new NullOutput());
+        }
+
+        $errIo = new SymfonyStyle($input, $output->getErrorOutput());
+        $errIo->setDecorated(!$input->getOption('no-ansi') || $input->getOption('ansi'));
+
+        return $errIo;
+    }
+}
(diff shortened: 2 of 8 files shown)


src/Command/Diff.php
```diff
@@ -0,0 +1,101 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $cwd = realpath($input->getArgument('cwd'));
+        $envType = EnvEnum::create($input->getOption('env'));
+
+        if (!is_dir($cwd)) {
+            throw new \RuntimeException(sprintf('The working directory "%s" doesn\'t exist.', $cwd));
+        }
+
+        $originalTemplatePath = MANALA_DIR.'/src/Resources/'.$envType;
```
chalasr: > A new EnvResourcesLocator might be useful for this. Or simply a new method on the EnvType.

Indeed 👍  some existing code and #29 will need it as well so let's add it in another PR

.php_cs
```diff
@@ -12,6 +12,7 @@ EOF;
 
 $finder = Symfony\CS\Finder\DefaultFinder::create()
     ->in(array(__DIR__.'/src', __DIR__.'/tests'))
+    ->exclude(array(__DIR__.'/tests/fixtures/'))
```
chalasr: no need for this change anymore :)

ogizanagi: May be required for other fixtures later...but as it doesn't work right now ¯\_(ツ)_/¯

chalasr:
Looks really nice!

ogizanagi:
This one is ready to me. PR description updated. Looking for reviews :)

However, I wonder if we shouldn't set a fake env (not usable with `manala setup`) in order to ease the test fixtures maintenance. WDYT ?

chalasr:
Imho, it is the right way.  I thought about `venv` when @xavier-rdo worked on #19, it might be useful for fully testing it. But maybe you have something else in mind?

ogizanagi:
@chalasr : I'm simply talking about something like this: https://github.com/manala/manala/pull/38/commits/7282583ca2a20a4504f191b69debcfbafe22c645

The `symfony` env will continue to evolve, but the `acme` one only has few reasons to do so and can act as a "sandbox" for our tests.
But we can probably found a mechanism to exclude it from normal usages, at least once the phar is build, until we have a dynamic way to manage envs.

chalasr:
Ok, sorry about the... :trollface: => 🚪 
For sure, it will widely ease maintaining it! 👍

.travis.yml
```diff
@@ -17,8 +18,10 @@ matrix:
 cache:
     directories:
         - $HOME/.composer/cache
-
```
chalasr, line 20: Could you please revert this change? :)

.travis.yml
```diff
@@ -17,8 +18,10 @@ matrix:
...
 before_install:
+    - sudo add-apt-repository -y ppa:git-core/ppa
+    - sudo apt-get update && sudo apt-get install git-man && sudo apt-get install -y git
+    - git --version
```
chalasr, line 25: Maybe this was for debugging purpose? Otherwise keeping it isn't a problem to me.

src/Exception/InvalidEnvException.php
```diff
@@ -18,9 +18,9 @@ class InvalidEnvException extends \InvalidArgumentException
     public function __construct($invalidEnvName)
     {
         parent::__construct(sprintf(
-            'The env "%s" doesn\'t exist. Possible values: [%s]',
+            'The env "%s" doesn\'t exist. Possible values: %s',
             $invalidEnvName,
-            implode(' ', EnvEnum::getPossibleEnvs())
+            json_encode(EnvEnum::getPossibleEnvs())
```
ogizanagi, line 23: ⚠️ Wild suggestion in the middle 😄

chalasr: 👍

src/Resources/acme/ansible/group_vars/app.yml
```diff
@@ -0,0 +1,131 @@
...
+  ##############
+  # Supervisor #
+  ##############
+
+  supervisor_configs:
+    - file:     app.conf
+      template: configs/app_program.{{ env }}.j2
+      config:
+        - workflow-worker:
+          - command:        php bin/console app:workflow:worker --run-max=1000 --shutdown-on-exception --no-interaction -vv
```
chalasr: Damn... This comes from dry*a, and not only this part is custom, meaning that the real template contains the same 😓, I have to fix it after merging this one

src/Env/EnvEnum.php
```diff
@@ -20,6 +20,7 @@
  */
 class EnvEnum
 {
+    const ACME = 'acme'; // This env is not meant to be used except for testing purposes.
```
chalasr: We should try to find a way to avoid making this env available outside of testing at time to fetch env templates remotely

tests/Functional/DiffTest.php
```diff
@@ -0,0 +1,39 @@
...
+namespace Manala\Tests\Functional;
+
+use Manala\Command\Diff;
+use Manala\Env\EnvEnum;
+use Symfony\Component\Console\Tester\CommandTester;
+
+class DiffTest extends \PHPUnit_Framework_TestCase
+{
+    const ACME_PROJECT_WD = MANALA_DIR.'/tests/fixtures/Command/DiffTest/AcmeProject';
+    const EXPECTED_OUTPUT_FILE = MANALA_DIR.'/tests/fixtures/Command/DiffTest/expected_acme.patch';
```
chalasr: A less expressive but simpler way to test this could be to create a temporary `symfony/framework-standard-edition` project (see [BuildTest](https://github.com/manala/manala/blob/master/tests/Functional/BuildTest.php#L38)), manalize it and edit some ansible files, giving an always predictible diff

src/Handler/Diff.php
```diff
@@ -0,0 +1,153 @@
...
+
+    private function copyToTmpLocation($templatePath)
+    {
+        $tmpPath = sys_get_temp_dir().'/Manala';
+
+        if ($this->fs->exists($tmpPath)) {
+            $this->fs->remove($tmpPath);
+        }
+
+        $this->fs->mkdir($tmpPath);
```
chalasr, line 137: Instead of using your own iterator, wouldn't `$this->fs->mirror($templatePath, $tmpPath);` work?

src/Command/Diff.php
```diff
@@ -0,0 +1,109 @@
...
+        if (!$handler->hasDiff()) {
+            $errIo->success('No diff found.');
+
+            return 0;
+        }
+
+        $errIo->note(<<<'EOTXT'
+You can apply the patch by executing the following command in the project working directory:
+
+$ patch -p1 < my_patch.patch
```
chalasr: How is set the patch filename? Is it `my_patch.patch`? I miss something ^^
Also imho we should be more explicit about that the project configuration is outdated, maybe something like:

```
Your project configuration is outdated, you can upgrade it by applying the patch on your project:

$ patch -p1 < my_patch.patch
```

chalasr (approved):
Left minor comments 👍

.travis.yml
ogizanagi: I left it in order to be sure there is no version with a git version later.

.travis.yml
ogizanagi: Sure

tests/Functional/DiffTest.php
ogizanagi: As soon as we always install the exact same version.

I thought about that, but I judge it was too much for the need and will slow down the test suite, and it doesn't solve the `acme` env issue (still required).

src/Handler/Diff.php
ogizanagi: That's probably the `Filesystem` method I haven't found. 😅 
Thanks, I'll give a try.

ogizanagi: Working great 👍 

Don't know how I missed this method :)

tests/Functional/DiffTest.php
chalasr: > As soon as we always install the exact same version.

Not sure to understand here, we manalize the temporary project with the same codebase as the one used to check the diff, so the pack is always up to date on both sides, right?

Also I find worst to have an unexisting env in our code base, but let's keep it (we have to quickly add the ability to load remote packs in order to remove it ^^)

src/Command/Diff.php
ogizanagi: > How is set the patch filename? Is it my_patch.patch? I miss something ^^

See https://github.com/manala/manala/pull/38/files#diff-45cbdbae2d21a8c8fae7d15a87e1b135R46

I finally choose to remove the mechanism I used to output the note on stderr, and simply added the text in the command help.

Thus, the command will only output the diff, and that's all. This also allow to use the command simply in quiet mode in order to get the exit code and know if there is a diff or not (0 or 1).

> we should be more explicit about that the project configuration is outdated

IMHO, having a diff is explicit enough.
I've only kept a `success` message to stderr in case there is no diff.

chalasr: Agreed, the diff directly given to the output is just perfect

ogizanagi:
@chalasr : Should we consider it ready and merge until we have the setup metadata feature ?
IMO it is already usable as it is.

chalasr:
@ogizanagi Great job 👍 Merged. Let's improve it later

PR #39, opened 2016-10-05
Add Diff command usage to README.md

ogizanagi:


chalasr:
Indeed, it's time to write to documentation 👍

PR #44, opened 2016-10-06
Avoid using Yaml for rendering ansible files

chalasr:
It was useless complexity. Btw the diff becomes far more acceptable

### DIFF
--- a/src/Env/Config/Renderer.php
+++ b/src/Env/Config/Renderer.php
@@ -11,9 +11,6 @@
 
 namespace Manala\Env\Config;
 
-use Manala\Env\Config\Variable\Variable;
-use Symfony\Component\Yaml\Yaml;
-
 /**
  * Config' template renderer.
  *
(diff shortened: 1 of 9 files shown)


ogizanagi:
So much better. looks good to me. 👍

ogizanagi (approved):

src/Env/Config/Renderer.php
```diff
@@ -40,45 +37,12 @@ public static function render(Config $config)
...
 
         $vars = $config->getVars();
-        $content = file_get_contents($template);
-
-        if ('yml' === $template->getExtension()) {
-            return self::renderYaml($content, $vars);
-        }
-
-        $rendered = $content;
+        $content = $rendered = file_get_contents($template);
```
ogizanagi:
``` diff
- $content = $rendered = file_get_contents($template);
+ $rendered = file_get_contents($template);
```

ogizanagi (changes requested):

src/Env/Config/Renderer.php
```diff
@@ -40,45 +37,12 @@ public static function render(Config $config)
...
-        $content = file_get_contents($template);
-
-        if ('yml' === $template->getExtension()) {
-            return self::renderYaml($content, $vars);
-        }
-
-        $rendered = $content;
+        $rendered = file_get_contents($template);
 
         foreach ($vars as $var) {
```
ogizanagi, line 42: We could use `array_reduce` in order to inline the `return` and avoid assignation. But this should be benchmarked. Nobody ain't time for that. :trollface:

ogizanagi (approved):

PR #45, opened 2016-10-06
Call make build from setup and deploy

chalasr:


src/Resources/symfony/ansible/group_vars/deploy_demo.yml
```diff
@@ -0,0 +1,19 @@
...
+  - web/app_test.php
+
+#########
+# Tasks #
+#########
+
+manala_deploy_tasks:
+  - make: install@demo
+  - make: build@demo
+  - service_restart: supervisor
```
ogizanagi: Is that required ? I don't have this in my project templates.

src/Resources/symfony/ansible/group_vars/deploy_demo.yml
```diff
@@ -0,0 +1,19 @@
+
+---
+
+###########
+# Removed #
+###########
+
+manala_deploy_removed:
+  - web/app_dev.php
```
ogizanagi: Should we remove this one by default on demo ? ping @nervo 

I'll personally be in favor of keeping it and only remove `app_test.php`

ogizanagi: For reference: https://github.com/Elao/symfony-standard/blob/master/ansible/group_vars/deploy_demo.yml

chalasr: I believe you're right, we do not remove it 👍

chalasr:
Done

ogizanagi (approved):

PR #47, opened 2016-10-06
Remove make targets

ogizanagi:
We do not need this feature anymore, as we won't use those commands elsewhere AFAIU (starting the provisioning remains the user responsibility and functional tests for environments will be aware of the tasks to execute).


chalasr:
👍  Nice catch

chalasr:
The diff is even lighter ❤️

PR #48, opened 2016-10-06
Resolve vagrant box version depending on dependencies version

### LINKED ISSUE #43: Use different versions of the vagrant box depending on the chosen PHP version
Sort as we can support PHP versions lower than 7

chalasr:
Allows manalizing php 5 projects (I'll add some build tests for them).
Closes #43

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
         "symfony/filesystem": "~3.1",
         "symfony/process": "~3.1",
         "symfony/stopwatch": "~3.1",
-        "composer/semver": "^1.4"
+        "composer/semver": "^1.4",
+        "nikic/iter": "^1.4"
     },
     "require-dev": {
         "symfony/console": "~3.2@dev",
--- a/src/Env/Config/Variable/VagrantBoxVersionResolver.php
+++ b/src/Env/Config/Variable/VagrantBoxVersionResolver.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of the Manala package.
+ *
+ * (c) Manala <contact@manala.io>
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Env\Config\Variable;
+
+use Manala\Env\Config\Variable\Dependency\Dependency;
+use Manala\Env\Config\Variable\Dependency\VersionBounded;
+use function iter\search;
+
+/**
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class VagrantBoxVersionResolver
+{
+    const DEFAULT_BOX_VERSION = '~> 3.0.0';
+
+    /**
+     * Resolves the vagrant box version.
+     *
+     * @param \Iterator $dependencies
+     *
+     * @return VagrantBoxVersion
+     */
+    public static function resolve(\Iterator $dependencies)
+    {
+        $php = self::getDependency('php', $dependencies);
+
+        if (null === $php) {
+            return new VagrantBoxVersion(self::DEFAULT_BOX_VERSION);
+        }
+
+        return new VagrantBoxVersion(self::resolveFromPhpVersion($php->getVersion()));
+    }
+
+    /**
+     * @return string
+     */
+    private static function resolveFromPhpVersion($phpVersion)
+    {
+        return (float) $phpVersion < 7 ? '~> 2.0.0' : self::DEFAULT_BOX_VERSION;
+    }
+
+    /**
+     * @param string   $name
+     * @param Iterator $dependencies
+     *
+     * @return VersionBounded|false
+     */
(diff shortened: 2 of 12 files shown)


composer.json
```diff
@@ -17,7 +17,8 @@
         "symfony/filesystem": "~3.1",
         "symfony/process": "~3.1",
         "symfony/stopwatch": "~3.1",
-        "composer/semver": "^1.4"
+        "composer/semver": "^1.4",
+        "nikic/iter": "^1.4"
```
chalasr, line 21: https://www.sitepoint.com/memory-performance-boosts-with-generators-and-nikiciter/

chalasr:
@nervo Here is the error output of the fail I have with `php 5.6` and `mysql 5.6` on `manala/app-dev-debian ~> 2.0.0`:

```
TASK [manala.php : extensions > Install packages] ******************************
failed: [localhost] (item=[u'php5-xdebug', u'php5-mysqlnd', u'php5-intl', u'php5-curl', u'php5-mbstring', u'php5-xml', u'php5-zip', u'php5-gd']) => {"failed": true, "item": ["php5-xdebug", "php5-mysqlnd", "php5-intl", "php5-curl", "php5-mbstring", "php5-xml", "php5-zip", "php5-gd"], "msg": "No package matching 'php5-mbstring' is available"}
```

I'm pretty sure you told me about this one, do you have a clue? Maybe a specific `mysql` version to use?

nervo:
That's a naughty one...
Php extensions are not 100% iso from versions to versions. You can work around this issue by updating the extensions array, but i would like to solve it on the role side.
Something like "hey, you want mbstring ? sure ? but i see you're on php 5.5, idiot. Ok, i won't crash, i will just move to th next. Thx, kiss"

chalasr:
The extension being really needed and installed anyway, to me the `php_extensions` option should reflect what the end user wants to see installed, thus it should contain `mbstring` if it is.

> i would like to solve it on the role side

I think it's legit, but would it be possible and is it worth it? 

I can always add a `PhpExtensionsResolver` similar to what is done here for the vagrant box version and keep in mind that when using php 5, `mbstring` is default installed even if not explicitly set in the `php_extensions` array

ogizanagi:
I'd rather say we should not care on the `manala/manala` side if the `mbstring` ext is shipped with the configured php version or not. The manala roles should simply ignore it. When upgrading to a newer php version, the `mbstring` extension will probably still be required and thus should be present in the manala roles configuration.

So 👍 for @nervo (but even if I do love the `hey, you want mbstring ? sure ? but i see you're on php 5.5, idiot. Ok, i won't crash, i will just move to the next. Thx, kiss` message suggestion and trashing users, I'd suggest simply moving to the next 😄 )

chalasr:
@ogizanagi That's what I meant by:

> The extension being really needed and installed anyway, to me the php_extensions option should reflect what the end user wants to see installed, thus it should contain mbstring if it is.

The workaround I proposed consists into do not add these extensions if they will fail, but in fact the extension will be installed. This option being called `php_extensions` and not `php_packages`, the fix should indeed be done on the role side which tries to install an unneeded package instead of just enabling an already installed one.

chalasr:
@nervo Failing extensions are `php5-mbstring`, `php5-xml` and `php5-zip` (enabled in `app_extensions` as `mbstring`, `xml` and `zip`). Removing them make the build pass with php `5.6`

tests/Functional/SetupTest.php
```diff
@@ -99,7 +152,29 @@ public function testExecuteWithDefaultAppNameUsesTheCurrentDirectoryName()
         $this->assertFileExists(self::$cwd.'/ansible/app.yml');
         $this->assertFileExists(self::$cwd.'/ansible/ansible.yml');
 
-        $this->assertContains(":name        => 'manala'",  file_get_contents(self::$cwd.'/Vagrantfile'));
+        $expectedDeps = <<<'YAML'
```
ogizanagi, line 203: No assertion on this ?

chalasr: oops 👍

src/Env/Config/Variable/VagrantBoxVersionResolver.php
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * @param string   $name
+     * @param Iterator $dependencies
+     *
+     * @return VersionBounded|false
+     */
+    private static function getDependency($name, \Iterator $dependencies)
+    {
+        return filter(function (Dependency $dependency) {
```
ogizanagi: Actually, I think it'll be better to use [search](https://github.com/nikic/iter/blob/master/src/iter.php#L601-L627). There should be no gain, but semantics. ¯\_(ツ)_/¯

chalasr: far better 👍  it avoids using `current()` at end, as it returns the found result directly

chalasr:
@ogizanagi All fixed

PR #49, opened 2016-10-06
Collect env vars into a ansible/.manala file

ogizanagi:
First step in order to keep metadata about the setup env, before reusing them in other commands, like the `diff` one for instance.

### DIFF
--- a/src/Env/Dumper.php
+++ b/src/Env/Dumper.php
@@ -13,6 +13,7 @@ namespace Manala\Env;
 
 use Manala\Env\Config\Renderer;
 use Symfony\Component\Filesystem\Filesystem;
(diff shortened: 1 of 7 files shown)


src/Env/Dumper.php
```diff
@@ -33,8 +34,15 @@ public static function dump(Env $env, $workDir)
     {
         $fs = new Filesystem();
 
+        $export = $env->export();
+        $fs->dumpFile("$workDir/ansible/.manala.yml", Yaml::dump([
+            'envs' => [
```
ogizanagi, line 39: JFYI: I've not handled the case (intentionally) in which we have a project using multiple environments, because it is not supported right now. But the metadata file structure needs to plan ahead this.

PR #52, opened 2016-10-08
Fix missing dependency when buildign the phar

ogizanagi:
Adding a new dependency in the project means adding it to the `box.json` file, unless the dep is only used for testing purposes.


chalasr:
https://github.com/manala/manala/commit/ead06d9d137b569de787976e9d262a546f94d27a :) Sorry for did it discretely

PR #54, opened 2016-10-09
Add php 7.1 to travis

chalasr:


.travis.yml
```diff
@@ -8,9 +8,10 @@ env:
 matrix:
     fast_finish: true
     include:
+        - php: 5.6
```
ogizanagi, line 11: For the record: I moved this one below in order for Travis to execute the fastest and most complete build fist (7.0). But actually, it doesn't seem the order has any effect :/

.travis.yml
```diff
@@ -8,9 +8,10 @@ env:
...
         - php: 7.0
           env: CHECK_PHP_SYNTAX="yes"
```
ogizanagi, line 13: There are chances the php 7.1 build is slightly faster than 7.0. I would move the `php-cs-fixer` check to this build :) WDYT ?

.travis.yml
chalasr: Yeah so I moved it up for readability

ogizanagi (approved):

.travis.yml
chalasr: Oops, merged :)

ogizanagi: 😄

PR #55, opened 2016-10-09
Simplify box.json & use compression

ogizanagi:
I think we don't need that ~~complexity~~ exhaustivity to reduce the phar size. It's a pain to keep the list of vendors updated in the `box.json` while we can already gain more by using compression.

With this, we've already gain ~500K (from 747K to 162K).

### DIFF
--- a/box.json
+++ b/box.json
@@ -1,45 +1,18 @@
 {
   "finder": [
+    { "in": "src" },
     {
-      "in": "src"
-    },
-    {
-      "name": "autoload.php",
+      "name": ["*.php", "LICENSE"],
+      "exclude": ["Tests", "Tester"],
       "in": "vendor"
-    },
-    {
-      "name": [
-        "autoload_real.php",
-        "ClassLoader.php",
-        "autoload_static.php"
-      ],
-      "in": [
-        "vendor/composer"
-      ]
-    },
-    {
-      "name": [
-        "*.php",
-        "LICENSE"
-      ],
-      "exclude": [
-        "Tests",
-        "Tester"
-      ],
-      "in": [
-        "vendor/symfony/console",
-        "vendor/symfony/filesystem",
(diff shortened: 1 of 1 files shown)


chalasr:
Could you btw the add the following patterns in `exclude`: `["Test", "test", "tests"]` ?
Some of them are actually not excluded

chalasr:
Arf, too late, I'll do it on master

ogizanagi:
https://github.com/manala/manalize/pull/54#discussion_r82518893

:trollface:

nervo:
'Careful on this one !
Composer has made the choice to _NOT_ use compression, as gzip extensions (or similar) are not always available by default.

ref: https://github.com/composer/composer/blob/master/src/Composer/Compiler.php#L155

ogizanagi:
@nervo : I'm not sure it can really be an issue. I don't see any mention about it on http://php.net/manual/en/phar.compress.php.

Would you have some references about this decision on Composer side ? :)

nervo:
:arrow_up:

ogizanagi:
@nervo : Ok. I also found some issues and PRs discussing about this on the composer repository.

But really, I do think we should not care about it. Our target is certainly not as wide as composer. Manalize will probably remain a tool only used on developers machines and will not be distributed in such kind of environments.

ogizanagi:
Also, for the record, the [symfony-installer](https://github.com/symfony/symfony-installer/blob/master/box.json#L22) uses it, and is certainly closer to our target than composer. :)

chalasr:
What kind of environment are concerned by this issue?

ogizanagi:
@chalasr : Mainly HHVM and systems without the gzip (zlib) extension.

nervo:
Symfony installer is a good point, let do this.

PR #56, opened 2016-10-09
Add `manala_get_tmp_dir()` function

ogizanagi:


bootstrap.php
```diff
@@ -3,4 +3,21 @@
...
 define('UPDATE_FIXTURES', filter_var(getenv('UPDATE_FIXTURES'), FILTER_VALIDATE_BOOLEAN));
+
+/**
+ * Creates a unique tmp dir.
+ *
+ * @param string $prefix
+ *
+ * @return string The path to the tmp dir created
+ */
+function manala_get_tmp_dir($prefix = '')
```
ogizanagi, line 16: Should it be `manalize_get_tmp_dir` and `define('MANALIZE_TMP_ROOT_DIR', sys_get_temp_dir().'/Manalize');` ?

chalasr: Imho it's ok as is

chalasr (approved):

chalasr:
Great! Thank you @ogizanagi

PR #57, opened 2016-10-10
Fix manala_get_tmp_dir parsing the whole tmp directory

chalasr:
I just realized that if the first argument passed to `tempnam()` is an unexisting directory, this one quietly use `sys_get_tmp_dir()` instead. So we need to first create the `MANALIZE_TMP_ROOT_DIR` before using `tempnam(MANALIZE_TMP_ROOT_DIR)`, then remove it in `*Test::tearDownAfterClass()`


ogizanagi:
😮  Didn't know that. Good catch 👍

PR #58, opened 2016-10-10
Move logic from Setup/SelfUpdate in handlers

chalasr:


src/Command/SelfUpdate.php
```diff
@@ -71,21 +72,33 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 0;
         }
 
+        $currentBuild = realpath($_SERVER['argv'][0]) ?: $_SERVER['argv'][0];
```
ogizanagi: Duplicated code with l.59 (if still relevant)

src/Command/SelfUpdate.php
```diff
@@ -55,10 +56,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
-        $this->client = new Client();
+        $handler = new SelfUpdateHandler(realpath($_SERVER['argv'][0]) ?: $_SERVER['argv'][0]);
```
ogizanagi, line 59: Seems it does not require this argument anymore.

src/Command/SelfUpdate.php
```diff
@@ -71,21 +72,33 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 0;
         }
 
+        $currentBuild = realpath($_SERVER['argv'][0]) ?: $_SERVER['argv'][0];
+
         try {
-            $this->updateBuild($latestTag);
-        } catch (\PharException $e) {
+            $handler->handle(function ($latestBuild) use ($currentBuild) {
```
ogizanagi: I wonder if the separation of concerns made here is relevant or if the `SelfUpdate` handler should both request the new phar and replace the old one. I don't see much value having this in the command which is only here to gather inputs from the outside env and user and display informations. This callback does none of these.

src/Handler/SelfUpdate.php
```diff
@@ -0,0 +1,75 @@
...
+namespace Manala\Manalize\Handler;
+
+use GuzzleHttp\Client;
+use Manala\Manalize\Application;
+use Manala\Manalize\Exception\HandlingFailureException;
+
+/**
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class SelfUpdate implements Handler
```
ogizanagi: I wonder if we really need this interface. Actually we're never using it but instead clearly rely on implementations (which is not something bad here).

chalasr: I think you're right

chalasr:
@ogizanagi `Handler` interface has been removed, this should be ready.

ogizanagi (approved):

ogizanagi:
Great !

PR #59, opened 2016-10-10
Add Test case

chalasr:
Depends on #58


tests/Handler/SetupTest.php
```diff
@@ -46,13 +46,9 @@ public function testHandle()
             $dumpedFiles[] = $target;
         });
 
-<<<<<<< HEAD
```
ogizanagi: Oops. Something wrong might have happened while rebasing.

PR #60, opened 2016-10-14
Update phony targets

ogizanagi:
As https://github.com/Elao/symfony-standard/pull/154

It avoid having any issue when a make target is named the same as a directory.


chalasr (approved):
👍

PR #61, opened 2016-10-19
Cleanup diff by re-building an env based on project's metadata

chalasr:


chalasr:
Diff best viewed [without whitespaces](https://github.com/manala/manalize/pull/61/files?w=1)

src/Env/Metadata/MetadataParser.php
```diff
@@ -28,9 +28,9 @@
      *
      * @return MetadataBag
      */
-    public static function parse(EnvEnum $envType)
+    public static function parse(EnvEnum $envType, $metadataPath = null)
```
ogizanagi: I think it's cleaner to have two methods:
- `parse(string $path)`
- `parseEnv(EnvType $envType)` in which is only build the path and called `parse(string $path)`

Also, `METADATA_PATH` may be renamed to `ENV_METADATA_PATH` or whatever, as there is two different metadata types.

src/Handler/Diff.php
```diff
@@ -126,13 +129,18 @@ public function getErrorOutput()
...
         $this->fs->mkdir($tmpPath);
-        $this->fs->mirror($templatePath, $tmpPath);
-        $this->fs->remove("$tmpPath/manala.yml");
+
+        $env = EnvFactory::createFromMetadata($this->envType, MetadataParser::parse($this->envType, $this->cwd.'/ansible/.manalize.yml'));
+        $dump = Dumper::dump($env, $tmpPath);
+
+        do {
+            $dump->next();
+        } while ($dump->valid());
```
ogizanagi:
``` php
for (; $dump->valid(); $dump->next());
```

or: 

``` php
foreach ($dump as $̄); // Yes it works
```

is way more fun 😄 

(See https://twitter.com/nicolasgrekas/status/586108028421804032 about the variable name in the foreach :D )

src/Env/Config/Variable/SingleValue.php
```diff
@@ -0,0 +1,50 @@
...
+
+    public static function createFromReplaces(array $replaces)
+    {
+        return new static($replaces[static::getPlaceholder()]);
+    }
+
+    /**
+     * @return string
+     */
+    abstract protected static function getPlaceholder();
```
ogizanagi: I don't know if we should really care, but according to http://php.net/manual/fr/migration52.incompatible.php : 

> Dropped abstract static class functions. Due to an oversight, PHP 5.0.x and 5.1.x allowed abstract static functions in classes. As of PHP 5.2.x, only interfaces can have them.

(I have the warning in PHPStorm). 
Maybe this should be moved in an interface ? :/

src/Env/Env.php
```diff
@@ -43,9 +44,21 @@ public function getConfigs()
...
     {
         $vars = [];
-        foreach ($this->getVars() as $var) {
-            foreach ($var->getReplaces() as $placeholder => $value) {
-                $vars[$placeholder] = $value;
+        $dependencies = [];
+
+        foreach ($this->getConfigs() as $config) {
+            foreach ($config->getVars() as $var) {
+                if ($var instanceof Dependency) {
```
ogizanagi: I'm not convinced the `Env` class should be aware of that :/

src/Handler/Diff.php
```diff
@@ -126,13 +129,18 @@ public function getErrorOutput()
...
+
+        $this->fs->remove("$tmpPath/ansible/.manalize.yml");
```
ogizanagi: Maybe we should split the env vs metadata dump in order to avoid that.

src/Handler/Diff.php
chalasr: `$̄` WTF? :joy:

src/Env/Dumper.php
```diff
@@ -25,21 +24,16 @@ class Dumper
...
     {
         $fs = new Filesystem();
 
-        $export = $env->export();
-        $fs->dumpFile("$workDir/ansible/.manalize.yml", Yaml::dump([
-            'envs' => [
-                $export['env'] => ['vars' => $export['vars']],
-            ],
-        ], 5));
+        $fs->dumpFile("$workDir/ansible/.manalize", serialize($env));
```
ogizanagi: You should still keeps the multi environments for the future, so this should at least be an array:

``` php
$fs->dumpFile("$workDir/ansible/.manalize", serialize([$env]));
```

Or maybe we don't care right now.

chalasr:
Thanks @ogizanagi!

PR #62, opened 2016-10-26
Allow to setup env metadata only

ogizanagi:
Sometimes, it can be handy to reconfigure the environment(s), without affecting the project files (in order to use the diff command with a new environment config for instance).

The following command will only update the `ansible/.manalize` metadata file:

```
$ manalize setup --no-update ~/my-awesome-project
```

(echoes the `composer require --no-upate` option)


ogizanagi:
Updated for PHP 5.6 😷  : 

``` diff
+if (self::DUMP_FILES & $flags) {
-    yield from self::dumpFiles($env, $workDir);
+    for ($generator = self::dumpFiles($env, $workDir); $generator->valid(); $generator->next()) {
+        yield $generator->current();
+    }
+}
```

src/Env/Dumper.php
```diff
@@ -21,15 +21,35 @@
...
      * @param string $workDir The manalized project directory
+     * @param int    $flags
      *
      * @return \Generator The dumped file paths
      */
-    public static function dump(Env $env, $workDir)
+    public static function dump(Env $env, $workDir, $flags = self::DUMP_ALL)
+    {
+        if (self::DUMP_FILES & $flags) {
+            // php >= 7.0:
```
chalasr: 😅

chalasr:
Let's wait for #68 for improving this :)

ogizanagi:
Updated

PR #67, opened 2016-10-27
Init the installer script

### LINKED ISSUE #65: Create an installer script
Some projects like [`box-project/box2`](https://github.com/box-project/box2) suggests to use an installer in order to automatically download the last version of the tool, set proper rights and move it to the binaries directory.

``` sh
# README.md
$ curl -L https://github.com/manala/manalize/releases/download/v0.2.1/manalize.phar > /usr/local/bin/manalize
$ chmod +x /usr/local/bin/manalize
```

On our side, it'll allow to forget about editing `README.md` at each new version, as we can do api calls on github in order to retrieve the last version automatically.

From the user's perspective, you'll only have to ~~type~~ copy/paste the following line:

``` sh
$ curl -LSs https://github.com/manala/manalize/blob/master/installer.php | php
```

We can also host the script on a dedicated repository and make it available from a more user-friendly uri, for instance: http://www.manala.io/installer.php

ogizanagi:
``` sh
# curl -LSs https://raw.githubusercontent.com/manala/manalize/master/installer.php | php
curl -LSs https://raw.githubusercontent.com/manala/manalize/installer/installer.php | php
```

capture d ecran 2016-10-27 a 14 23 50
capture d ecran 2016-10-27 a 14 24 11
capture d ecran 2016-10-27 a 14 24 19

(fixes #65 )

### DIFF
--- a/installer.php
+++ b/installer.php
@@ -1,0 +1,213 @@
+<?php
+
+/*
+ * This file is part of the Manalize project.
+ *
+ * (c) Manala <contact@manala.io>
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Manalize\Installer;
+
+class Installer
+{
+    const DEFAULT_DIR = '/usr/local/bin';
+
+    private $repository;
+    private $io;
+
+    public function __construct()
+    {
+        $this->repository = new ManalizeRepository();
+        $this->io = new IO();
+    }
+
+    public function install($dir = self::DEFAULT_DIR)
+    {
+        $io = $this->io;
+        try {
+            $io->newLine();
+            $io->title('Manalize Installer');
+
+            if ('\\' === DIRECTORY_SEPARATOR) {
+                throw new \RuntimeException('Sorry, but Windows is not supported right now');
+            }
+
+            if (!is_dir($dir) || !is_writeable($dir)) {
+                throw new \RuntimeException(sprintf(
+                    'Target directory at path "%s" should exist and be writable',
+                    $dir
+                ));
+            }
+
+            $io->newLine();
+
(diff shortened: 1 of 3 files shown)


chalasr:
Is this testable? :)

Really nice work! 👍

installer.php
```diff
@@ -0,0 +1,211 @@
...
+                '10.0.10586' === PHP_WINDOWS_VERSION_MAJOR.'.'.PHP_WINDOWS_VERSION_MINOR.'.'.PHP_WINDOWS_VERSION_BUILD
+                || false !== getenv('ANSICON')
+                || 'ON' === getenv('ConEmuANSI')
+                || 'xterm' === getenv('TERM');
+        }
+
+        return function_exists('posix_isatty') && @posix_isatty(STDOUT);
+    }
+}
+
```
chalasr, line 195: I think we should check the PHP version here

ogizanagi:
> Is this testable? :)

I don't know. Maybe by testing the `Installer` class by adding `installer.php` to the classmap in tests, or some kind of integration test by executing `curl -LSs https://raw.githubusercontent.com/manala/manalize/installer/installer.php | php` in a process, but I should find a way to change the target path for the executable in order to put it in a tmp dir.

installer.php
ogizanagi: Finally, I won't. We talked about showing how to use manalize through docker for those who don't want to install another php version locally. But they still need to install manalize on their host, so we can't restrict the script.

chalasr: Hmm.. now I have some stuff in mind I would like to discuss, let's see that at office.

ogizanagi:
@chalasr : Finally got some time to make this little test !

This is ready.

chalasr:
Great, thank you @ogizanagi

PR #68, opened 2016-10-27
Drop php 5.6 support

### LINKED ISSUE #63: Consider moving to PHP 7
- According to [PHP's Roadmap](http://php.net/supported-versions.php), PHP 5.6 will be actively supported only up to 1st January 2017.
- Until Manalize is spread _all around the world_, PHP 7 will be the norm.
- Manalize is a tool used by developers, on their own machines. It should not be an issue, as any respectful developer should have the latest PHP version installed. :p 
- We're depriving ourself from using some really cool features (see https://github.com/manala/manalize/pull/62#issuecomment-256407154 for instance)
- Really ? :trollface:

chalasr:
Closes #63


src/Command/Setup.php
```diff
@@ -75,7 +75,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         return 0;
     }
 
-    private function setupDependencies(SymfonyStyle $io, MetadataBag $metadata)
+    private function setupDependencies(SymfonyStyle $io, MetadataBag $metadata) : \Generator
```
ogizanagi: There is no convention yet about this AFAIK, but I personally prefer:

``` diff
- private function setupDependencies(SymfonyStyle $io, MetadataBag $metadata) : \Generator
+ private function setupDependencies(SymfonyStyle $io, MetadataBag $metadata): \Generator
```

(no space between the closing parentheses and the colon)

WDYT ?

chalasr: I kept the space from what I can see in some popular repositories, but I'm looking for a standard... Do you have such?

ogizanagi: I based my choice on https://wiki.php.net/rfc/scalar_type_hints
But as I said, I don't know any consensus.

ogizanagi: See https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2207/files#diff-88b99bb28683bd5b7e3a204826ead112R523

chalasr: Quick win! 👍

ogizanagi: This fixer is not released yet however 😢  It's only for 2.0.

chalasr: I thought about moving to 2.x-dev within this PR. Wdyt?

ogizanagi: 👍

nervo:
Hm, we still have, and still support some projects in 5.4, 5.5 and 5.6

ogizanagi:
@nervo : It's not about dropping the support of PHP 5.6 in virtualized environments, but on the host on which is used the manalize tool in order to init a project.

nervo:
Did'nt we see that osx (even sierra release) is still in 5.6 ?

ogizanagi:
That's entirely true. However, see https://github.com/manala/manalize/issues/63 :)

nervo:
Hm, ok.
:half_convinced: :)

ogizanagi:
Let's fight tomorrow about that ☕️ 🚬 :D

ogizanagi:
As I did it for formageddon, I added 1ed26b5 to switch to php-cs-fixer v2 ;)

chalasr:
Great, thank you @ogizanagi!

@nervo I understand your pov, I personally had some doubts about this, thinking to people like frontend developers who just do not want to hear about PHP.

But we already have some requirements (which are not pre-installed on osx/linux), so PHP 7 would just be another one, and probably the easier to install

```
$ curl -s https://php-osx.liip.ch/install.sh | bash -s 7.0
```

(if we prefer homebrew, we just have to publish a shell script executing [the 5 commands it needs](https://developerjack.com/blog/2015/12/11/Installing-PHP7-with-homebrew/))

Imho, it is worth it.

ogizanagi:
Let's give @nervo the responsibility for merging (_or closing_) this one (symbolism, U know :trollface: )

ogizanagi:
Thanks @chalasr ! :)

PR #69, opened 2016-10-28
Remove some client's default values

ArnaudBenassy:
- [x] Remove client's default values 
- [x] Add var/sessions as symbolic link
- [x] Stay up-to-date with symfony-standard

### DIFF
--- a/src/Resources/symfony/ansible/group_vars/app.yml
+++ b/src/Resources/symfony/ansible/group_vars/app.yml
@@ -19,13 +19,6 @@ app_options:
 
 app_patterns:
 
-  ########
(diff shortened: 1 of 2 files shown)


chalasr:
👍 Thanks. We just need to execute the build test locally (which runs make setup and is excluded from travis builds) in order to verify if we don't need to create the session directory if it doesn't exist.

ogizanagi:
The test case is fine 👍 

Thank you @ArnaudBenassy

chalasr:
Thanks @ArnaudBenassy @ogizanagi

PR #70, opened 2016-10-28
Switch to manala skeleton

nervo:
Ok, sit down, have a cigar, and listen carefuly.

Right now, templates (or call them like you want) are not vendored. I mean "symfony" is just "symfony". As we propose (at least) two skeletons "manala_app_php" and "elao_app_php", we must begin to think about it.

### DIFF
--- a/src/Resources/symfony/ansible/app.yml
+++ b/src/Resources/symfony/ansible/app.yml
@@ -5,7 +5,7 @@
     # App
     app: "{{ manala_skeleton_options.app }}"
     # Manala
-    manala_skeleton_name: elao_app_php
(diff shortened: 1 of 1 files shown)


ArnaudBenassy (approved):

chalasr:
@nervo could you please briefly remind me the difference between the two skeletons?

ogizanagi:
@chalasr : That's mainly about the motd & nginx error templates.

But we need to talk about more things on Monday... :'(

ogizanagi:
Thank you @nervo

chalasr:
👍 Thanks!

PR #71, opened 2016-10-30
Propose default dependency set before asking for custom

chalasr:
As quickly discussed, this adds a question proposing to keep the default dependency set of the env instead of configuring them one by one.

Output (may be improved):

### DIFF
--- a/src/Handler/Setup.php
+++ b/src/Handler/Setup.php
@@ -12,9 +12,12 @@
 namespace Manala\Manalize\Handler;
 
 use Manala\Manalize\Env\Config\Variable\AppName;
+use Manala\Manalize\Env\Config\Variable\Dependency\Dependency;
+use Manala\Manalize\Env\Config\Variable\Dependency\VersionBounded;
 use Manala\Manalize\Env\Dumper;
 use Manala\Manalize\Env\EnvEnum;
 use Manala\Manalize\Env\EnvFactory;
+use Manala\Manalize\Env\Metadata\MetadataBag;
 
 /**
  * @author Robin Chalas <robin.chalas@gmail.com>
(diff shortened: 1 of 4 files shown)


src/Handler/Setup.php
```diff
@@ -27,13 +30,8 @@ class Setup
...
     private $options;
 
-    public function __construct(
-        string $cwd,
-        AppName $appName,
-        EnvEnum $envName,
-        \Traversable $dependencies,
-        array $options = []
-    ) {
+    public function __construct(string $cwd, AppName $appName, EnvEnum $envName, \Traversable $dependencies, array $options = [])
```
ogizanagi: Should we chop down methods' arguments if too long, or not ? (We're constantly changing this 😝 )

ogizanagi (approved):
I like it 👍

src/Handler/Setup.php
chalasr: I agree! I think we have to define what does mean "too long" 😅  Currently, I line-break as soon as the signature is larger than my 15" macbook pro, which is quite stupid as all the context is took in consideration i.e. font size, screen resolution, tree view's width... 

Imho it's ok with large signatures. The only argument favoring line-breaking I have is about debugging, when the problem comes from one of the args present in the signature, the line shown in the trace always corresponds to the failing arg so there is often no need for looking deeper into the trace.

What is your preference? I guess you're more a line break kind of guy :)

ogizanagi: @chalasr : Too long is as always: 120 chars long.
If you use PHPStorm, you can configure it to do it automatically for you.

It's not much about debugging for me, only about visual pollution, and the fact we've chosen this convention at Elao (at least we did for GBA, now I apply it on every projects).

ogizanagi: Here you can see both PHPStorm (through method line indicators) and code sniffer detected the line is too long.

capture d ecran 2016-10-31 a 11 47 39

ogizanagi:
Thank you @chalasr

PR #73, opened 2016-10-30
Add ScriptHandler for manalizing project using composer events

chalasr:
So we can just remove the post-create-project hook from `elao/symfony-standard` and use `Manala\\Manalize\\Composer\\ScriptHandler::manalizeProject` instead.

### DIFF
--- a/src/Composer/ScriptHandler.php
+++ b/src/Composer/ScriptHandler.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Manalize project.
+ *
+ * (c) Manala <contact@manala.io>
+ *
+ * For the full copyright and license information, please refer to the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Manala\Manalize\Composer;
+
+use Symfony\Component\Console\Application as BaseApplication;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\ConsoleOutput;
+use Manala\Manalize\Command\Setup;
+
+/**
+ * Manalize composer script handler.
+ *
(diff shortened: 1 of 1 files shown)


ogizanagi:
Hmmm... This is meant to replace the `post-create-project-cmd` hook script from the `elao/symfony-standard`, but isn't the aim of manalize to forget about the `elao/symfony-standard` edition (and remove it in the "long" term)?

chalasr:
I think that make the `elao/symfony-standard` uses this project can be a good first step. Then we could see, in the long run, if the framework edition can be totally removed.

Also, making the `elao/symfony-standard` dependent of this project at first can help I think, giving it some visibility. 
Let's discuss this @elao :)

Btw, I thought about adding a `ScriptHandler::updateProject()` that would run `diff`, warn user if there is a patch and maybe propose to apply it. WDYT?

ogizanagi:
I'm not against this change, but I think it's currently too idealistic. But indeed, let's talk about it at the office :)

> Btw, I thought about adding a ScriptHandler::updateProject() that would run diff, warn user if there is a patch and maybe propose to apply it. WDYT?

There'll always be a warning due to how it works right now 😢  It can be quite annoying.

chalasr:
> I think it's currently too idealistic.

~~What do you mean? I tried it and it works well, except that there is some things we don't handle right now, like the `behat.yml.dist`, I'll soon add it.~~ indeed

> There'll always be a warning due to how it works right now 

The diff runs successful if there is no change remotely, what would cause a permanent warning?

chalasr:
Let's burninate the elao standard edition :)

nervo:
Well well well :)

PR #74, opened 2016-11-06
Extract/Hydrate Envs for creating/reusing metadata

chalasr:
Using basic serialization is too much restrictive, any small refactoring would make impossible to run `manalize diff` thus be a BC break. 

Here is a new approach which makes use of reflection for Variable objects data extraction and lambdas for hydration, giving such metadata:

```yaml
symfony:
    vagrant:
        app_name:
            - { value: foo-bar.manala }
        box_version:
            - { value: '~> 3.0.0' }
    ansible:
        dependency_with_version:
            - { version: '7.0', name: php, enabled: true }
            - { version: '5.6', name: mysql, enabled: true }
            - { version: '9.5', name: postgresql, enabled: false }
            - { version: '3.2', name: mongodb, enabled: false }
            - { version: '1.7', name: elasticsearch, enabled: false }
            - { version: '6', name: nodejs, enabled: false }
        dependency:
            - { name: redis, enabled: false }
            - { name: influxdb, enabled: false }
    make: {  }

```

I didn't benched yet but executing the test suite locally gives no significant difference in term of execution time (the whole test suite takes more or less 1 more second to finish) and makes no difference at all for memory usage.


src/Env/EnvFactory.php
```diff
@@ -24,13 +25,36 @@
...
+
+            foreach ($perClassVars as $class => $vars) {
+                foreach ($vars as $data) {
+                    $var = (new \ReflectionClass($class))->newInstanceWithoutConstructor();
+                    $hydrator->hydrate($var, $data);
+                    $hydratedVars[] = $var;
+                }
+            }
+
+            $configs[] = new $configClass($name, ...$hydratedVars);
```
ogizanagi: Should probably define a named constructor instead, otherwise there is nothing guaranteeing the constructor looks like this, right ? (even if we own that code, I think it's a clearer path)

ogizanagi (approved):
Alright, we can try this way.
It's still better than raw serialization 👍. 
However I don't think it'll help much more about refactoring, because you still use the FQCNs.

You could renamed the `*.txt` fixtures inside `tests/fixtures/Command/SetupTest/` to `*.yml` :)

src/Env/Defaults/Defaults.php
```diff
@@ -9,27 +9,27 @@
...
+namespace Manala\Manalize\Env\Defaults;
 
 /**
- * A metadata bag containing all defaults for a given env type.
+ * A bag containing all defaults for a given env.
  *
  * @author Robin Chalas <robin.chalas@gmail.com>
  */
-final class MetadataBag
+final class Defaults
```
ogizanagi, line 19: I'm not sure I like much this renaming.
What are your motivations behind this ?

chalasr: The motivation is that we need to make a clear distinction between the project's metadata and the env defaults, it's too much confusing to have the same naming for both. About the choice of `defaults`, it just sounds logical to me as the file present in the env contains its defaults. I'm quite surprised by your comment but let's see if we can find a better one or another way to make this less confusing, it started to become harder to keep this separation clear enough.

chalasr:
> However I don't think it'll help much more about refactoring, because you still use the FQCNs.

Sure... I thought about adding a `Config::getName()` and iterate over all config classes until finding the matching one and your comment makes me think it is finally worth it

src/Env/EnvFactory.php
chalasr: Good point, I think we need to first check if it is a subclass of Config (if it isn't, throw an exception) and either make `Config::__construct()` final or define a named constructor which would be final too, right?

src/Env/Defaults/Defaults.php
ogizanagi: Nevermind. I was actually confused between the actual metadata, defaults, and this class.
After pulling the code, this totally make sense... 👍 😆

src/Env/EnvFactory.php
ogizanagi: > make Config::__construct() final or define a named constructor which would be final too, right?

Making the `__construct` final could be an option. But I would avoid it, as it'll prevent doing anything in the constructor in a child class if we need it in the future.

The named constructor doesn't need to be final, as it has to respect exactly the signature (that's the only thing we need) and still allows to override the implementation. It can still define extra arguments with defaults, but I don't think we should care.

chalasr: Correct. 👍

ogizanagi:
> Sure... I thought about adding a Config::getName() and iterate over all config classes until finding the matching one and your comment makes me think it is finally worth it

Looks like you'll need some kind of registry for accessing all config classes and get their name (no need for the `Config::getName()` method then).

chalasr:
@ogizanagi Looks like php-cs-fixer already made breaking changes since you set it up to `@dev`. Maybe we should require a specific commit until a version is released?

ogizanagi:
@chalasr They've released a beta yesterday. We should probably stick to it. What about `2.x@beta` ?

ogizanagi:
Also the `2.x@dev` constraint should have been enough, but we don't have the [`prefer-stable: true` flag inside the `composer.json` file](https://github.com/Elao/PhpEnums/blob/master/composer.json#L26) (don't know if we should).

chalasr:
This needs a review 🙄

tests/Env/Config/Variable/VariableExtractorTest.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+namespace Manala\Manalize\Tests\Env\Config\Variable;
+
+use Manala\Manalize\Env\Config\Variable\AppName;
+use Manala\Manalize\Env\Config\Variable\VariableExtractor;
+
+class VariableExtractorTest extends \PHPUnit_Framework_TestCase
+{
+    public function testExtracte()
```
ogizanagi:
``` diff
- testExtracte
+ testExtract
```

ogizanagi (approved):
Looks good to me appart from minor comment

ogizanagi:
Great job!

PR #75, opened 2016-11-06
CS Fixer: Stick to beta

ogizanagi:
Or should we prefer the `prefer-stable` option for more safety ?


chalasr:
Thanks @ogizanagi

ogizanagi:
@chalasr : You didn't answer the question 😅 

(my bad, I've added it after creating the PR)

chalasr:
Indeed, I missed it! Let's keep it in mind :)

PR #76, opened 2016-11-07
Don't try to run manalize on PHP_VERSION < 7

ogizanagi:


nervo:
wow, such reactivity

chalasr:
Thank you @ogizanagi

PR #78, opened 2016-11-10
Update available dependencies

chalasr:


src/Resources/symfony/manala.yml
```diff
@@ -5,7 +5,7 @@ packages:
         default:    '7.0'
         constraint: '5.4|5.5|5.6|7.0'
     mysql:
-        enabled:    true
+        enabled:    false
```
ogizanagi: Not enabled by default anymore?

chalasr: It isn't on [the upstream](https://github.com/Elao/symfony-standard/blob/master/ansible/group_vars/app.yml#L8) and I agree, node is  not enabled by default here, I think the minimal should be proposed. WDYT?

ogizanagi: Alright 👍

chalasr: `SetupTest` fixtures need to be updated due to this change...

ogizanagi: Also needs a rebase 😅

PR #80, opened 2016-11-11
Write to stderr when relevant

chalasr:


src/Command/Diff.php
```diff
@@ -109,7 +109,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
     private function getErrIo(InputInterface $input, OutputInterface $output)
```
ogizanagi, line 110: What about making this method a helper and use it at other places in the code?

ogizanagi (commented):
As I said, I don't know if it has much importance for those commands. It should be considered mainly when the command is mixing both outputs. For instance, the only means of the diff command is to output the generated patch, so it makes sense to output anything else to the err output.
For setup and selfupdate, there is no real output desired for these commands, so you won't redirect or pipe the stream with this intent.

But if we're going this way, I'm not sure

``` php
$output instanceof ConsoleOutputInterface ? $output->getErrorOutput() : $output
```

would be right (see the helper instead). As soon as you want to output something on stderr, it should not fallback to stdout. That'll go against the purpose of these changes, isn't it?

chalasr:
Indeed, it's not worth it since the changed commands produce no relevant output.

PR #83, opened 2016-11-17
Update php default extensions

nervo:
Gd should not be installed as a default extension


chalasr:
Nice catch, thanks @nervo

PR #84, opened 2016-11-19
Fix process error catching in diff

chalasr:
Given `git diff` exits with `1` for both failure (e.g. unknown command) and success revealing some diff, we can't refer to the exit code for differentiating them.

Before:
before

After:
after

ping @ogizanagi

### DIFF
--- a/src/Command/Diff.php
+++ b/src/Command/Diff.php
@@ -85,20 +85,18 @@ EOTXT
             throw new \RuntimeException(sprintf('The working directory "%s" doesn\'t exist.', $cwd));
         }
 
+        $errIo = $this->getErrIo($input, $output);
         $handler = new DiffHandler($envName, $cwd, $output->isDecorated());
 
         try {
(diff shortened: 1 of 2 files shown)


src/Command/Diff.php
```diff
@@ -85,21 +85,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         try {
-            $handler->handle(function ($type, $buffer) use ($output) {
-                $output->write($buffer);
+            $handler->handle(function ($stdout) use ($output) {
+                $output->write($stdout);
             });
         } catch (HandlingFailureException $e) {
-            $io = new SymfonyStyle($input, $output);
-            $io->error(['An error occurred during the process execution:', $handler->getErrorOutput()]);
+            $errIo->error(['An error occurred during the process execution:', $handler->getErrorOutput()]);
```
chalasr, line 96: Used the error io btw, so we can safely rely on stdout even in case of fail.

PR #85, opened 2016-11-19
Ask for strategy on already existing file in setup

### LINKED ISSUE #64: Consider using git diff & apply in order to setup & update envs
Replacing the `Dumper` class by applying a patch on the project may help supporting multi-envs.
Only the diff about what is missing for the project to comply the targeted env will be generated.

It may help merging existing files with necessary changes brung by the env resources (Makefile, .gitignore, ...).
If the patch cannot be applied due to conflict, we may suggest the user to create the patch and apply it interactively in order to resolve them.

The `Dumper` will only be kept in order to dump manalize metadata (we may also keep it in order to dump the env with resolved placeholders in a tmp directory though).

Drawbacks: it can only be applied on a git initialized repository (or we should fallback to the `patch` util).

chalasr:
The `setup` command actually erases any already existing file in the project if it corresponds to one of the environment configuration files, i.e. if the manalized project already has a Makefile, it will just be replaced by the env's one.

Since it is common to have a Makefile in an existing project (or an ansible related file if the project was running under ElaoInfra for instance), I propose to let the user choose between 3 strategies:
- Replace it
- Give me a patch
- Do nothing

Output for an already existing Vagrantfile:

out

If the patch strategy is chosen, then the end user can do ~~`git apply Vagrantfile.patch`~~ `git apply manalize.patch`.
Tests will be added depending on feedbacks.

- [x] Add tests

Closes #64 (different approach)

### DIFF
--- a/src/Command/Diff.php
+++ b/src/Command/Diff.php
@@ -85,27 +85,25 @@ EOTXT
             throw new \RuntimeException(sprintf('The working directory "%s" doesn\'t exist.', $cwd));
         }
 
-        $errIo = $this->getErrIo($input, $output);
         $handler = new DiffHandler($envName, $cwd, $output->isDecorated());
+        $errorIo = $this->getErrorIo($input, $output);
 
         try {
             $handler->handle(function ($diff) use ($output) {
                 $output->write($diff);
+            }, function () use ($errorIo) {
+                $errorIo->success('No diff found.');
             });
         } catch (HandlingFailureException $e) {
-            $errIo->error(['An error occurred during the process execution:', $handler->getErrorOutput()]);
+            $errorIo->error(['An error occurred during the process execution:', $e->getMessage()]);
 
-            return $handler->getExitCode();
+            return 1;
         }
 
-        if (!$handler->hasDiff()) {
-            $errIo->success('No diff found.');
-        }
-
-        return $handler->getExitCode();
+        return 0;
     }
 
-    private function getErrIo(InputInterface $input, OutputInterface $output)
+    private function getErrorIo(InputInterface $input, OutputInterface $output)
     {
         if (!$output instanceof ConsoleOutput) {
             return new SymfonyStyle($input, new NullOutput());
--- a/src/Command/Setup.php
+++ b/src/Command/Setup.php
@@ -18,6 +18,7 @@ use Manala\Manalize\Env\Defaults\Defaults;
 use Manala\Manalize\Env\Defaults\DefaultsParser;
 use Manala\Manalize\Env\Dumper;
 use Manala\Manalize\Env\EnvName;
+use Manala\Manalize\Exception\HandlingFailureException;
 use Manala\Manalize\Handler\Setup as SetupHandler;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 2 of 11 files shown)


src/Command/Diff.php
```diff
@@ -86,25 +86,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
         }
 
         $handler = new DiffHandler($envName, $cwd, $output->isDecorated());
+        $errorIo = $this->getErrIo($input, $output);
```
ogizanagi: Minor but I'd suggest `$errIo`

src/Command/Setup.php
```diff
@@ -125,4 +127,34 @@ private function shouldConfigureDependencies(SymfonyStyle $io, Defaults $default
 
         return $io->confirm('Do you want to customize your dependencies?', false);
     }
+
+    private function askStrategyForExistingFile(SymfonyStyle $io, string $cwd, string $target, array $strategies): string
+    {
+        $strategy = $io->choice(
+            sprintf('The file "%s" needs to be updated.', str_replace($cwd.'/', '', $target)),
```
ogizanagi, line 134:
```diff
- The file "%s" needs to be updated.
+ The file "%s" conflicts with suggested changes.
```

?

src/Command/Setup.php
```diff
@@ -125,4 +127,34 @@ private function shouldConfigureDependencies(SymfonyStyle $io, Defaults $default
...
+            array_values($strategies),
+            $strategies[Dumper::DO_REPLACE]
```
ogizanagi: I suggest to make the patch the default (or nothing, but not replace).

src/Handler/Setup.php
```diff
@@ -44,12 +45,16 @@ public function __construct(
...
     {
         $env = EnvFactory::createEnv($this->envName, $this->appName, $this->dependencies);
 
-        foreach (Dumper::dump($env, $this->cwd, $this->getDumperFlags()) as $target) {
-            $notifier(str_replace($this->cwd.'/', '', $target));
+        try {
+            foreach (Dumper::dump($env, $this->cwd, $this->getDumperFlags(), $existingFileCallback) as $target) {
+                $notifier(str_replace($this->cwd.'/', '', $target));
+            }
+        } catch (\RuntimeException $e) {
```
ogizanagi, line 58: `FailedDumpException` ?

ogizanagi (approved):

src/Handler/Setup.php
chalasr: I did this so both `FailedDumpException` and (filesystem) `IOException` are caught

src/Command/Diff.php
chalasr: What's the motivation behind that? Since stderr is just a shortcut for "standard error output" (even if it's quite misleading) I don't think that shorten `error` to `err` would make it better

ogizanagi: Why did you name the method `getErrIo` then? :trollface:

chalasr: But... that's [your code](https://github.com/manala/manalize/commit/ef3dfb68dbc0c2781208b144a164cb5e40f76b57#diff-45cbdbae2d21a8c8fae7d15a87e1b135R99) 😅

ogizanagi: Sure. But I expected you to update it if the naming was not conform to your own expectations ^^

chalasr: Ah.. Because it stays good enough for an internal so I thought it's not worth it. But if you think it makes sense I can update it (I would not want to hurt you :trollface:)

src/Command/Setup.php
chalasr: At this point we're not sure if there is any conflict, but only that we need to update this file because it already exists. Maybe you were thinking that I was trying to apply the patch before asking this question?

chalasr:
Do we consider that this closes #64?

ogizanagi:
I'd say yes, because it solves the issue with existing files with a slightly different approach as the one described (#64 would have eliminated the `Dumper` class (except for metadata)). We may reconsider using git as the one core util for this later if we encounter another issue, but for now this looks like the best of both worlds 😃

ogizanagi:
`vendor/bin/php-cs-fixer fix` and ship it? :shipit:

PR #88, opened 2016-11-24
Enlarge symfony deps version bound

chalasr:


composer.json
```diff
@@ -15,11 +15,10 @@
     "minimum-stability": "dev",
     "require": {
         "php": ">=7.0",
-        "symfony/console": "^3.1",
-        "symfony/filesystem": "^3.1",
-        "symfony/process": "^3.1",
-        "symfony/stopwatch": "^3.1",
```
chalasr, line 21: removed stopwatch btw (see https://github.com/manala/manalize/pull/87#discussion_r89519358)

ogizanagi:
I've almost thought of a spam :trollface: (enlarge your p...rojects)

ogizanagi:
Did you tested it with lowest dependencies? We should update travis too.

chalasr:
> Did you tested it with lowest dependencies?

Yes it works :)

> We should update travis too

Indeed!

ogizanagi:
Also, why do we still require `"symfony/console": "~3.2@dev"` in dev?

chalasr:
For the awesome `CommandTester#setInputs()` :trollface: :)

ogizanagi:
TOTALLY WORTH IT 👍 . Sorry. 😄

ogizanagi:
(I've added the `In Progress` label for keeping in mind we need to update the `.travis.yml` file to test on lowest dependencies too in the matrix.)

.travis.yml
```diff
@@ -4,13 +4,18 @@ language: php
...
 matrix:
     fast_finish: true
     include:
         - php: 7.0
         - php: 7.1
           env: CHECK_PHP_SYNTAX="yes"
+        - php: 7.0
+          env: DEPS="low"
+        - php: 7.1
+          env: DEPS="low"
```
ogizanagi: I think only one build with `env: DEPS="low"` is enough actually.

PR #89, opened 2016-11-26
Make use of enums

chalasr:


composer.json
```diff
@@ -22,7 +22,8 @@
         "symfony/yaml": "^3.1",
         "composer/semver": "^1.4",
         "nikic/iter": "^1.4",
-        "guzzlehttp/guzzle": "^6.2"
+        "guzzlehttp/guzzle": "^6.2",
+        "elao/enum": "^0.4.0"
```
ogizanagi, line 26: Should use the tilde operator until a 1.0.0 release is out, because <1.0.0 the BC policy is only for patches.
Thus, minor versions can break the api until we consider it stable.

tests/Env/DumperTest.php
```diff
@@ -86,7 +86,7 @@ private function createEnv()
             ->willReturn(null);
 
         $env = EnvFactory::createEnv(
-            EnvEnum::create(EnvEnum::SYMFONY),
+            EnvName::SYMFONY(),
```
ogizanagi, line 89: I'd recommend to add the methods in the enum types docblock if we use this notation, as described in https://github.com/Elao/PhpEnums#shortcuts to keep static code analysis and autocompletion viable :)

src/Env/EnvName.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        return [
+            self::SYMFONY,
+        ];
+    }
+
+    public static function readables(): array
+    {
+        return [
+            self::SYMFONY => self::SYMFONY,
```
ogizanagi:
```php
self::SYMFONY => 'Symfony',
```

?

src/Exception/InvalidEnvException.php
```diff
@@ -20,7 +20,7 @@ public function __construct($invalidEnvName)
         parent::__construct(sprintf(
             'The env "%s" doesn\'t exist. Possible values: %s',
             $invalidEnvName,
-            json_encode(EnvEnum::getPossibleEnvs())
+            json_encode(EnvName::getPossibleEnvs())
```
ogizanagi:
```diff
- json_encode(EnvName::getPossibleEnvs())
+ json_encode(EnvName::values())
```

src/Requirement/Requirement.php
```diff
@@ -16,31 +16,36 @@
...
  * @author Xavier Roldo <xavier.roldo@elao.com>
  */
-class Requirement implements Common\RequirementLevelHolderInterface
+class Requirement
 {
-    use Common\RequirementLevelHolderTrait;
-
-    const TYPE_BINARY = 'binary';
-    const TYPE_VAGRANT_PLUGIN = 'vagrant_plugin';
+    // use Common\RequirementLevelHolderTrait;
```
ogizanagi: to remove?

ogizanagi (changes requested):

composer.json
ogizanagi: Erratum:  

> For pre-1.0 versions it also acts with safety in mind and treats ^0.3 as >=0.3.0 <0.4.0.

I've forgotten about this ^^

tests/Env/DumperTest.php
chalasr: done! 👍

ogizanagi (approved):

PR #90, opened 2016-11-26
Add git to host recommendations

chalasr:


README.md
```diff
@@ -31,6 +31,7 @@ Prerequisites
 -------------
 
 - [PHP](http://php.net) 7.0+
+- [GIT](https://git-scm.com/) 2.6+
```
ogizanagi: For the record: Why 2.6+?

chalasr: Because it's my current version 😆 😄

ogizanagi: OMG. You're late :trollface: 

```sh
$ git --version
git version 2.9.3
```

https://github.com/git/git/releases

Not sure it worth be as accurate then. Git 2+ should be enough (until someone reports any issue justifying it).

nervo:
Eeeeeeasy, boys :) https://packages.debian.org/jessie/git

ogizanagi:
@nervo : So 2+ is good then 👅 (Anyway required for the diff command to work properly)

ogizanagi:
But actually, it's not a strong requirement right now (only used for diff).

chalasr:
> only used for diff

Hum, `setup` will require it in #85

ogizanagi:
I know 😄 I said **right now**, and we don't have to make it required in order to the setup command to work (as for the deploy role and ansible, it can be considered as a bonus. So a warning could simply advised the user it's required for such cases). We're not relying on git as a core tool to make the project setup as described in https://github.com/manala/manalize/issues/64.

chalasr:
Ok you win

chalasr:
Good to go!

chalasr:
Now it's good to go 😄

PR #103, opened 2016-12-21
Mysql should be default enabled for the symfony env

chalasr:
/cc @cmeiller


ogizanagi:
Thanks for him and everyone 👍

PR #105, opened 2016-12-22
Fail deploy on unreachable hosts

nervo:


chalasr:
For the record, I fixed the tests using `UPDATE_FIXTURES=1 vendor/bin/phpunit --exclude-group infra`.

Thank you @nervo, you're awesome.

PR #113, opened 2017-03-21
Hype

chalasr:
- [x] Update template rendering/dumping logic according to the new directory structure (`manala/`)
- [x] dump a writable `manala.yml` including the equivalent of the current `app_options` (must continue to dump `app_options` in `group_vars/app.yml` until roles get able to read from `manala.yml`)
- [x] Dump `manala/metadata.yml` with some hard-coded values (see `manalize-templates/elao-symfony/manala/metadata.yml
- [x] Update `diff` accordingly
- [x] Fix tests
- [x]  READ-ONLY vars (app_dir, ...)
- [ ] Manifest validation/deep merging
- [ ] Restrict diff to `manala/*` directory
- [ ] Change from `manalize diff` to `manalize update` with `--diff` 
- [ ] Rename `manala` to `.manala`


src/Command/Setup.php
```diff
@@ -73,7 +73,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         try {
             $syncer = new Syncer();
-            $syncer->sync();
+            $syncer->sync('4be52bb');
```
chalasr: testing purpose; must be reverted before merge

PR #114, opened 2017-05-19
Don't dump read-only vars in a dedicated file, use twig instead

chalasr:


tests/fixtures/Command/SetupTest/app_1.yml
```diff
@@ -20,6 +21,17 @@ app_options:
...
+# App #
+#######
+
+app:
+    name:      "manalized-app"
+    user:      "app"
+    dir:       "/srv/app"
+    log_dir:   "/var/log/app"
+    cache_dir: "/var/cache/app"
+
```
chalasr, line 34: @nervo OK for you?

PR #119, opened 2017-06-20
Setup command: select environment name from a choice list

xavier-rdo:
Fixes https://github.com/manala/manalize/issues/115

- [x] Add tests

### DIFF
--- a/src/Command/Setup.php
+++ b/src/Command/Setup.php
@@ -47,7 +47,7 @@ class Setup extends Command
             ->setName('setup')
             ->setDescription('Configures your environment on top of Manala ansible roles')
             ->addArgument('cwd', InputArgument::OPTIONAL, 'The path of the application for which to setup the environment', getcwd())
-            ->addOption('env', null, InputOption::VALUE_OPTIONAL, 'One of the supported environment types. Don\'t use this option for building a full custom environment', null)
+            ->addOption('env', null, InputOption::VALUE_OPTIONAL, sprintf('One of the supported environment types: [%s]', implode(', ', EnvName::values())), null)
             ->addOption('no-update', null, InputOption::VALUE_NONE, 'If set, will only update metadata')
         ;
     }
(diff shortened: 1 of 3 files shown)


src/Command/Setup.php
```diff
@@ -139,6 +140,14 @@ private function getEnvName(InputInterface $input, string $cwd): EnvName
             return EnvName::get($templateName);
         }
 
+        if ($envName = $this->getEnvNameFromChoiceList($io)) {
+            try {
+                return EnvName::get($envName);
+            } catch (InvalidValueException $exception) {
+                return EnvName::CUSTOM();
```
ogizanagi: Tu as la possibilité de set un validateur en 3ème argument de `SymfonyStyle::ask()` afin de répéter la question dans le cas d'une typo par exemple, plutôt que de fallback automatiquement sur le custom.

chalasr: He uses `SymfonyStyle::choice()` which has a [default validator](https://github.com/symfony/console/blob/master/Question/ChoiceQuestion.php#L127), dunno how it behaves but maybe it re-asks naturally? Anyway re-asking seems better than going with `CUSTOM` in case of bad answer

src/Command/Setup.php
```diff
@@ -139,6 +140,10 @@ private function getEnvName(InputInterface $input, string $cwd): EnvName
             return EnvName::get($templateName);
         }
 
+        if ($envName = $this->getEnvNameFromChoiceList($io)) {
+            return EnvName::get($envName);
+        }
```
chalasr: couldn't it be `return $this->getEnvNameFromChoiceList();`, removing the `CUSTOM` fallback? Actually the fallback can't be reached since the question will be repeated until a valid answer is given

chalasr:
👍  Thanks @xavier-rdo

xavier-rdo:
Thanx to you for sharing your knowledge, Robin !

PR #121, opened 2017-06-23
[Doc] some minor updates in the main README

xavier-rdo:
- [x] Some minor updates in README
- [x] Add a "how to contribute" part in README

**Note:** I introduce a new sentence in the "_What's inside_"  part : 

```
Manalize is built on the shoulders of the following libs :
```

Manalize is a tool that provides other tools (a _scaffolder_ in some way) and "what's inside" seems ambiguous to me : what's inside manalize or what's inside what manalize provides ? ;) Adding this sentence makes it less ambiguous according to me.

Resolves https://github.com/manala/manalize/issues/120

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,25 +3,27 @@ Manalize
 
 [![Build Status](https://travis-ci.org/manala/manalize.svg?branch=master)](https://travis-ci.org/manala/manalize)
 
-This project provides ready-to-use environments for various projects.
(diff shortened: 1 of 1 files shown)


chalasr:
I agree `What's inside` is misleading, in fact I thought about removing it (manalize isn't an extension point, if one wants to know more about the binary itself then one could just open its composer.json). Let's keep it, the sentence you added make the difference.

Thanks @xavier-rdo!

PR #124, opened 2017-08-29
Fix Blackfire config keys in app_local.yml.sample templates

ogizanagi:
⚠️ This reverts commits that can't belong to a 0.5.x patch. Do not merge as is!

It aims to fix the blackfire config (https://github.com/manala/manalize/commit/5201c5d999af45d9ccb76d63d53578d11a662912) and provide a patch release for 0.5 version.

As quickly discussed on Gitter, I think the 0.6 version deserves its own branch until ready to get merged on master and be released :)


chalasr (approved):
I just pushed a `0.5.x` branch in sync with master, going to merge this as is into, then up to master where I'd reintroduce the reverted changes as new commits. Ok for you?

ogizanagi:
Looks good to me, but then we could even just merge this as is, tag & release 0.5.4, and reintroduce reverted commits. It would be the same, right?
The separate 0.6 branch would allow to keep being able to create patch versions if we find other issues.
Or keep master as is, merge this in 0.5.x, and keep the branch forever :/

chalasr:
Another alternative would be merge this as is in master and re-add reverted commits in #113, not add version-specific branches and only merge self-sufficient features. WDYT?

ogizanagi:
👍 :shipit:

ogizanagi:
Just added the cs fix BTW, so master will be green.

chalasr:
Thanks!

PR #127, opened 2017-10-10
Fix manual build instructions

voronkovich:
It's needed to run `composer install` (to install the `box`) before calling `make build`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,9 @@ $ composer global require manala/manalize
 
 #### Using git:
 ```
-$ git clone git@github.com:manala/manalize
+$ git clone https://github.com/manala/manalize
 $ cd manalize
+$ composer install
 $ make build
 $ mv manalize.phar /usr/local/bin/manalize
 $ chmod a+x /usr/local/bin/manalize
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ $ composer global require manala/manalize
 
 #### Using git:
 ```
-$ git clone https://github.com/manala/manalize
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,8 +52,9 @@ $ composer global require manala/manalize
 
 #### Using git:
 ```
-$ git clone git@github.com:manala/manalize
+$ git clone https://github.com/manala/manalize
```
ogizanagi: Why this change?

voronkovich: @ogizanagi, Because git asks a password when `git@github.com:manala/manalize` is used.

ogizanagi: You probably miss your ssh key then, because `git@github.com:manala/manalize` is using SSH.

Using HTTPS, it also asks for your credentials, but you probably did enter them only once if you use a credential helper.

See: 
- https://help.github.com/articles/why-is-git-always-asking-for-my-password/
- https://help.github.com/articles/caching-your-github-password-in-git/

For what it's worth, I'd still recommend using SSH, however it seems [Github recommends using HTTPS](https://help.github.com/articles/which-remote-url-should-i-use/), even if [they changed many times their recommendations](https://stackoverflow.com/a/11041782)...

A good sum up of the different ways to access github repos: https://gist.github.com/grawity/4392747

So, why not.

voronkovich: @ogizanagi, Thanks for the explanation! I've reverted the `git` protocol!

ogizanagi:
Good catch, thanks @voronkovich.

PR #132, opened 2017-11-06
Fix sqlite missing from app_options && enable by default

ogizanagi:
Sqlite was missing from the `app_options` template.
Also, I'd suggest to enable it by default, as it's convenient for small apps to start without the VM and have a continuous evolve path when switching to it. But if I'd enable it by default, that's because it's also useful for tests with in memory database, which improves perfs for functional tests: https://github.com/Elao/climalife-k2/pull/970


chalasr (approved):
👍  Nice

chalasr:
Thank you @ogizanagi.

PR #139, opened 2017-12-12
Add PHP 7.2

chalasr:


NicolasDievart:
Could add an entry here ? https://github.com/manala/manalize/blob/master/tests/Env/Config/Variable/VagrantBoxVersionResolverTest.php#L24

And the setup tests ?

And maybe .travis.yml too ?

chalasr:
> Could add an entry here ? https://github.com/manala/manalize/blob/master/tests/Env/Config/Variable/VagrantBoxVersionResolverTest.php#L24

nice catch! done

> And the setup tests ?

done already, see updated fixtures

> And maybe .travis.yml too ?

unrelated but would have to be done anyways, done.
Thanks @NicolasDievart

ogizanagi (approved):

PR #141, opened 2017-12-13
Replace TLD .dev by .vm

ogizanagi:
Change the tld `.dev` by `.vm`. We'll eventually provide a feature to change it interactively when running `manalize setup`.

Fixes part of https://github.com/manala/manalize/issues/137


NicolasDievart:
@ogizanagi No vendor option then ?

ogizanagi:
That's only a quick-win until we provide a better feature to choose it.
It'll have an impact on the cli command, metadata and need to introduce a new var in the Vagrantfile template.

chalasr (approved):
self sufficient

chalasr:
Thank you @ogizanagi.

PR #145, opened 2018-04-30
Update README with prerequisite of PHP 7.1

NicolasDievart:
We check that php 7.1 is installed to use manalize (https://github.com/manala/manalize/pull/143) but the readme still says it is ok with php 7.0+


ogizanagi:
Thank you @NicolasDievart.

PR #150, opened 2018-07-11
add elastic search 6 support

mb3rnard:


src/Resources/envs/symfony/defaults.yml
```diff
@@ -37,7 +37,7 @@ packages:
         enabled:    false
         required:   false
         default:    '5'
```
chalasr: should we make it the default version?

mb3rnard: I don't see why we couldn't :) I've made the change

chalasr:
Fixtures needs to be updated :)
`UPDATE_FIXTURES=1 make test` should fix them all

chalasr (approved):

chalasr:
Thank you @mb3rnard.

PR #152, opened 2018-10-19
Install "unzip" APT package to prevent problems with Composer

Kocal:
Some weeks ago we encountered [an issue](https://github.com/symfony/panther/issues/54) when installing [symfony/panther](https://github.com/symfony/panther) in our VM because `unzip` was missing.

The execution bit was missing on a binary file packaged in Panther, which makes us not able to run Panther (without `chmod +x` the binary, OFC).

@dunglas found the issue (https://github.com/symfony/panther/issues/54#issuecomment-427972572), it's because package `unzip` wasn't installed, so it's use the buggy PHP Zip extension that is not able to keep the executable bit of binaries...

So I think it would be nice to always have `unzip` installed to prevent some problems.

### DIFF
--- a/src/Resources/envs/custom/ansible/group_vars/app.yml
+++ b/src/Resources/envs/custom/ansible/group_vars/app.yml
@@ -4,6 +4,13 @@
 
 app_patterns:
 
+  #######
+  # Apt #
+  #######
+
+  apt_packages:
+    - unzip
+
   ############
   # Timezone #
   ############
--- a/src/Resources/envs/symfony/ansible/group_vars/app.yml
+++ b/src/Resources/envs/symfony/ansible/group_vars/app.yml
@@ -4,6 +4,13 @@
 
 app_patterns:
 
+  #######
+  # Apt #
+  #######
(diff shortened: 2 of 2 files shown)


ogizanagi:
I agree. But I think @nervo may have something else in mind. Perhaps it should be part of [manala.skeleton](https://github.com/manala/ansible-roles/blob/737d47fe2accbb90ead24680b5d8793f7ce5d79d/manala.skeleton/vars/main.yml#L442-L453) instead?

Kocal:
Oh, tests are failing.
Should I fix them or we are waiting for @nervo?

BTW I agree, I think it would be better if `unzip` is part of manala.skeleton.

ogizanagi:
Let's wait for @nervo 's opinion.

nervo:
Good question...
What about in `manala.composer` role ?

nervo:
Here it is: https://github.com/manala/ansible-roles/pull/252

Kocal:
Thank you

PR #155, opened 2018-11-19
Fix facts gathering on ansible playbook

### LINKED ISSUE #153: manalize uses gather_facts: false
When we use manalize, it still generates playbooks with `gather_facts: false` when it shouldn't be present (or set to `true` to avoid re-provisionning errors)

nervo:
Due to some distros limitations (yes, i'm looking at you, debian wheezy) with advanced ssl, `manala.ansible-galaxy` role must now gather facts to decide if it has to disable certificates checking or not.

Fix #153

### DIFF
--- a/src/Resources/envs/custom/ansible/ansible.yml
+++ b/src/Resources/envs/custom/ansible/ansible.yml
@@ -2,7 +2,6 @@
 
 # App
 - hosts: app
-  gather_facts: false
   vars:
     manala_ansible_galaxy_roles:
       - manala.ansible-galaxy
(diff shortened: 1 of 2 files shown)


ogizanagi:
There are some more: https://github.com/manala/manalize/search?q=gather_facts&unscoped_q=gather_facts

tests will fail since the expectations aren't updated :)

nervo:
Ooops, you're right, i forgot 1 commit somewhere...

ogizanagi (approved):

ogizanagi:
Thank you @nervo.

PR #157, opened 2018-11-19
Fix some deps

ogizanagi:
Fixing some deps so it works even on deps=low.

We shoudl consider switching from `kherge/box` to https://github.com/humbug/box

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ env:
     global:
         - CHECK_PHP_SYNTAX="no"
         - DEPS="high"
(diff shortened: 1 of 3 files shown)


nervo (approved):

chalasr (approved):
remaining failure seems related to https://github.com/symfony/symfony/pull/29263

ogizanagi:
Indeed. Let's force to Phpunit 6.5 for now?

chalasr:
Works for me

PR #158, opened 2018-11-19
Update environment variables handling

nervo:
Using `environment_variables` is deprecated, in favor of `app.env_vars` which handle properly environment variables both in zsh *and* php fpm

See: https://github.com/manala/ansible-role-skeleton/commit/9afb906b6da4d8a3ad2d285acea72699ddc5eeaa


ogizanagi:
Thank you @nervo.

PR #159, opened 2018-11-19
Remove php zip extension

nervo:
'was needed by default for composer only. Composer now prefers using system unzip binary, which is now installed by related role.

See: https://github.com/manala/ansible-role-composer/commit/ab84be2a24caa4797578853a887c258fa7b349ea


chalasr (approved):

ogizanagi:
Thank you @nervo.

PR #167, opened 2019-07-09
Hyphen gate

greedybro:
see https://github.com/manala/ansible-roles/pull/351


ogizanagi:
Thank you @greedybro.

